PR #3, opened 2022-09-09
Networking Setup for Application and Database Connectivity

manavrajvanshi-nx:
Creates a Service and an Endpoints object that maps to the provisioned database on NDB. This provides a mechanism for application pods running within the K8s cluster to connect to the database via the service (and wait for the database to be provisioned before starting up using init-containers if required). 

`make test`

`None`

**Release note**:

```release-note
1. Once the Database custom resource has reached the 'READY' state:
    a. A Service and an Endpoints object with the name DATABASE_CR_NAME-svc are created.
    b. The Endpoints object maps to the IP Address of the database provisioned on NDB.
2. Replaced some hardcoded strings with constants.
3. Added tests.
4. Updated README.
```


matthewfindlay (approved):

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -186,19 +186,32 @@ func GetOOBProfiles(ctx context.Context, ndbclient *ndbclient.NDBClient, dbType
...
+func GetDatabasePortByType(dbType string) int32 {
+	switch dbType {
+	case DATABASE_TYPE_POSTGRES:
+		return DATABASE_DEFAULT_PORT_POSTGRES
+	case DATABASE_TYPE_MONGODB:
+		return DATABASE_DEFAULT_PORT_MONGODB
+	case DATABASE_TYPE_MYSQL:
+		return DATABASE_DEFAULT_PORT_MYSQL
+	default:
+		return 80
```
deepak-nutanix, line 211: why not 443?

controllers/database_reconciler_helpers.go
```diff
@@ -253,3 +258,124 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+	}
+	log.Info("Returning from database_reconciler_helpers.setupService")
+	return
+}
+
+// Checks and creates an endpoints object for the service if it does not already exists.
+// If it is already present, syncs the IP address with the Database.Status.IPAddress if out of sync.
+func (r *DatabaseReconciler) setupEndpoints(ctx context.Context, database *ndbv1alpha1.Database, namespacedName types.NamespacedName, metadata metav1.ObjectMeta, targetPort int32) (err error) {
+	log := ctrllog.FromContext(ctx)
+	log.Info("Entered database_reconciler_helpers.setupEndpoints")
```
deepak-nutanix, line 336: Rather than this type of comments, would it make sense to say that you are starting to set endpoints for a given database, under a given namespace, etc

controllers/database_reconciler_helpers.go
```diff
@@ -253,3 +258,124 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+		}
+		log.Info("Endpoint found with a different IP address, updating.")
+		foundEndpoint.Subsets = endpointSubsets
+		err = r.Update(ctx, foundEndpoint)
+		if err != nil {
+			log.Error(err, "Failed to update endpoint")
+			return
+		}
+	}
+	log.Info("Returning from database_reconciler_helpers.setupEndpoints")
```
deepak-nutanix, line 379: Rather than this type of comments, would it make sense to say that you are were able to successfully were able to set endpoints for a given database, under a given namespace, etc

controllers/database_reconciler_helpers.go
```diff
@@ -253,3 +258,124 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+					// IP has not changed, no need to update endpoint
+					return
+				}
+			}
+		}
+		log.Info("Endpoint found with a different IP address, updating.")
+		foundEndpoint.Subsets = endpointSubsets
+		err = r.Update(ctx, foundEndpoint)
+		if err != nil {
+			log.Error(err, "Failed to update endpoint")
```
deepak-nutanix, line 375: Can you add more context?

controllers/database_reconciler_helpers.go
```diff
@@ -253,3 +258,124 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+		log.Info("No endpoint found, creating a new endpoint")
+		endpoint := &corev1.Endpoints{
+			ObjectMeta: metadata,
+			Subsets:    endpointSubsets,
+		}
+		// Setting database as the owner of this endpoint
+		ctrl.SetControllerReference(database, endpoint, r.Scheme)
+		err = r.Create(ctx, endpoint)
+		if err != nil {
+			log.Error(err, "Failed to create a new ep")
```
deepak-nutanix, line 356: rather than short forms, it would be great if you say endpoint.

deepak-nutanix (approved):
Minor comments. LGTM

api/v1alpha1/ndb_api_helpers.go
manavrajvanshi-nx: This could ideally be anything, I used 80 as a placeholder for the default block. Will not reach a condition where the 'default' port will be used.

controllers/database_reconciler_helpers.go
manavrajvanshi-nx: Not really, this method will be executed on every reconcile request (even after the endpoints have been set in an earlier reconcile). In that case, this method won't create / set endpoint. However, it will check if the endpoints have been changed and log an appropriate message later.

controllers/database_reconciler_helpers.go
deepak-nutanix: Ohh! What I'm trying to convey is that it would always be better to add logical log statements rather than saying that we are entering method x and exiting method x. This works really well when we are trying to debug. But, adding log statements would make more sense.

controllers/database_reconciler_helpers.go
manavrajvanshi-nx: An [endpoint subset](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoints-v1/) contains all the mappings of IPs and Ports for the Endpoints object. 
As already commented in the code :
image
if the IP address of the Database CR is same as the IP address of the database on NDB, the function returns from here.

If not, this means that the endpointSubset is out of sync (IPs are different), then the function updates the endpointSubsets field of the found endpoint as shown here:
image

image

controllers/database_reconciler_helpers.go
manavrajvanshi-nx: Yes, they are printed **only** when it is created.
image

PR #25, opened 2022-09-27
15 fix code scanning alert disabled tls certificate check

### LINKED ISSUE #15: Fix code scanning alert - Disabled TLS certificate check
Tracking issue for:
- [ ] https://github.com/nutanix-cloud-native/ndb-operator/security/code-scanning/1

manavrajvanshi-nx:

- Adds the ability to provide a SSL certificate for communication with the NDB server.
- Adds the ability to skip the SSL verification by a boolean flag in the CR yaml.
- Replaces plain text credentials with Kubernetes Secrets.

**Which issue(s) this PR fixes**:
Fixes #15 

Unit tests done through `make test`.
SSL Certificate tests done manually.

**Release note**:

```release-note
NONE
```


util/get_data_from_secret.go
```diff
@@ -0,0 +1,31 @@
...
+		log.Error(err, "Error occured while trying to read secret", "Secret Name", name, "Secret Namespace", namespace)
+	} else {
+		if val, ok := secret.Data[key]; ok {
+			data = string(val)
+			log.Info("Returning from util.GetDataFromSecret")
+		} else {
+			err = fmt.Errorf(fmt.Sprintf("Key '%s' not present in the secret '%s' in namespace '%s'", key, name, namespace))
+		}
+	}
+	return
```
matthewfindlay, line 30: IS this function supposed to return something?

matthewfindlay (approved):

util/get_data_from_secret.go
manavrajvanshi-nx: Yes, this function returns the 'data' and 'err'. These are declared/named in the function signature's return values. 
This is called named/naked return in golang. 
https://github.com/golang/go/wiki/CodeReviewComments#naked-returns
https://go.dev/tour/basics/7

matthewfindlay: cool!

util/get_data_from_secret.go
```diff
@@ -0,0 +1,31 @@
+package util
+
+import (
```
deepak-nutanix, line 3: would it make sense to rename the file to "secret.go" rather than get_data_from_secret.go ? You can add additional utility functions under this file.

Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 0.0.2
```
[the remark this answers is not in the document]
manavrajvanshi-nx: We have changed the CRD of the Database type and will need to create a new release. Hence the change in the version.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -50,6 +50,31 @@ func GenerateProvisioningRequest(ctx context.Context, ndbclient *ndbclient.NDBCl
 
 	database_names := strings.Join(dbSpec.Instance.DatabaseNames, ",")
 
+	// Type assertion
+	dbPassword, ok := reqData[NDB_PARAM_PASSWORD].(string)
```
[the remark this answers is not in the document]
manavrajvanshi-nx: I feel like password pattern matching and other limitations should not be the responsibility of the operator. If the provisioning request fails then that should be returned in the response by NDB, which is handled by the handlySync method.
image

controllers/database_controller.go
```diff
@@ -68,11 +68,21 @@ func (r *DatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
-	spec := database.Spec
-	server := spec.NDB
-	ndbClient := ndbclient.NewNDBClient(server.Credentials.LoginUser, server.Credentials.Password, server.Server)
-
-	// log.Info(fmt.Sprintf("Finalizers: %v", database.Finalizers))
+	NDBInfo := database.Spec.NDB
+	secretName := NDBInfo.CredentialSecret
+	username, err := util.GetDataFromSecret(ctx, r.Client, secretName, req.Namespace, ndbv1alpha1.SECRET_DATA_KEY_USERNAME)
+	if err != nil {
+		log.Error(err, "Error reading username from secret", "Secret Name", secretName)
```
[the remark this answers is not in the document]
manavrajvanshi-nx: Thanks, that's a good catch, I'll address this by returning a re-queue with an error.

util/get_data_from_secret.go
manavrajvanshi-nx: Makes sense!

controllers/database_controller_test.go
```diff
@@ -25,6 +25,8 @@ import (
 	"context"
 	"time"
 
+	b64 "encoding/base64"
```
deepak-nutanix, line 28: Would it be possible to cover more negative test cases. For instance, can we check the negative test scenario that all the errors are caught?

controllers/database_controller.go
```diff
@@ -68,11 +68,23 @@ func (r *DatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 	log.Info("Database CR Status: " + util.ToString(database.Status))
 
-	spec := database.Spec
-	server := spec.NDB
-	ndbClient := ndbclient.NewNDBClient(server.Credentials.LoginUser, server.Credentials.Password, server.Server)
-
-	// log.Info(fmt.Sprintf("Finalizers: %v", database.Finalizers))
+	NDBInfo := database.Spec.NDB
+	secretName := NDBInfo.CredentialSecret
+	username, err := util.GetDataFromSecret(ctx, r.Client, secretName, req.Namespace, ndbv1alpha1.SECRET_DATA_KEY_USERNAME)
```
deepak-nutanix, line 73: Rather than doing this, do you think we can add one function that would extract all the necessary information in one stretch? 
For instance, we are extracting information one by one: first username, then password, then ca-cert. Seems like this can be moved to a single function.

deepak-nutanix (commented):
Sorry, missed these comments.

deepak-nutanix (dismissed):
If possible, try to address the above comments or else the rest of the code looks good to me.

controllers/database_controller_test.go
manavrajvanshi-nx: Since it is not straightforward to have a functioning mock infrastructure (mocked NDB server + mocked Kubernetes cluster + communication between these mocks) it is not a feasible option to test this controller/reconcile logic (cannot inject mocks for various cases). However, I have added negative test cases for the pieces of codes / helper methods that this controller calls for the expected error cases.

controllers/database_controller.go
manavrajvanshi-nx: Extracted the function to return a map of the secret data and created a helper method to fetch these credentials from the map in one go. This was a great idea, it reduced the K8s API call to 1 from 3 in every reconcile function call. 😄

deepak-nutanix (dismissed):

README.md
```diff
@@ -29,12 +29,39 @@ make deploy
...
-1. To create instances of custom resources (provision databases), edit [ndb_v1alpha1_database.yaml](config/samples/ndb_v1alpha1_database.yaml) file with the NDB installation and database instance details and run:
+1. Create the secrets that are to be used by the custom resource:
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: your-ndb-secret
+type: Opaque
+data:
+  username: username-encoded-in-base64
```
tuxtof, line 40: let's move username and password in the stringData section

README.md
```diff
@@ -29,12 +29,39 @@ make deploy
...
+    -----BEGIN CERTIFICATE-----
+    CA CERTIFICATE (ca_certificate is optional)
+    -----END CERTIFICATE-----
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: your-db-secret
+type: Opaque
+data:
```
tuxtof, line 53: same here exemple is always better in stringData

tuxtof (approved):

matthewfindlay (approved):

PR #51, opened 2023-01-05
bump version to 0.0.3

tuxtof:

bump version to 0.0.3

**Release note**:

```release-note
NONE
```


manavrajvanshi:
Hi @tuxtof, Thanks for bumping the version, my understanding is that this version bump is required to create a new build of the operator. But can you please help me understand why do we have an 'ignore-for-release' label added in this PR? Shouldn't we release the bumped version?

manavrajvanshi (approved):

tuxtof:
> Hi @tuxtof, Thanks for bumping the version, my understanding is that this version bump is required to create a new build of the operator. But can you please help me understand why do we have an 'ignore-for-release' label added in this PR? Shouldn't we release the bumped version?

the goal is jsut to avoid to see `bump version to 0.0.3` in the release notes of the 0.0.3 
because it sounds logic :-D

krunal-jhaveri (approved):

PR #87, opened 2023-04-28
Feature/enhancements

tilaksatra:
This PR is a fresh one & includes simplified function & callbacks.
Moreover, I used only the util filter function that was provided and need not feel the need to create another one.

However, there is one point that I couldn't find alternative to:

customProfileOptions := [...]string{PROFILE_TYPE_COMPUTE, PROFILE_TYPE_SOFTWARE, PROFILE_TYPE_NETWORK, PROFILE_TYPE_DATABASE_PARAMETER}
	for _, profileValue := range customProfileOptions {
		err = MatchAndSetProfile(ctx, profileValue, profiles.Compute, allProfiles, profileMap)
		if err != nil {
			return
		}
	}
	
In the above call, I had to resort to for loop for profileType as I couldn't find a better way to do error handling:
- "Panic" would cause the program to fail & hence avoided it.
- Secondly, just calling MatchAndSetProfile for each profile Type would involve making err check for each of those call which would be repeated code.
- Any particular approach could be discussed.

Note: All Unit Tests passing.


DESIGN_DOCUMENT.md
manavrajvanshi, line 1: Not to be checked in

api/v1alpha1/database_types.go
```diff
@@ -208,12 +208,19 @@ type TaskInfoSummaryResponse struct {
...
+	EngineType      string    `json:"engineType"`
+	LatestVersionId string    `json:"latestVersionId"`
+	Topology        string    `json:"topology"`
+	Versions        []Version `json:"versions"`
+}
+
+type Version struct {
+	Id          string `json:"id"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
```
manavrajvanshi, line 223: We don't require the version anymore. We'll resolve the profiles either by name and/or by id, giving preference to name if both are provided.

api/v1alpha1/database_types.go
```diff
@@ -208,12 +208,19 @@ type TaskInfoSummaryResponse struct {
...
-	Topology        string `json:"topology"`
+	Id              string    `json:"id"`
+	Name            string    `json:"name"`
+	Type            string    `json:"type"`
+	EngineType      string    `json:"engineType"`
+	LatestVersionId string    `json:"latestVersionId"`
+	Topology        string    `json:"topology"`
+	Versions        []Version `json:"versions"`
+}
+
```
manavrajvanshi, line 219: We might not want to run code-gen for this.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -159,18 +159,68 @@ func GetNoneTimeMachineSLA(ctx context.Context, ndbclient *ndbclient.NDBClient)
...
+	allProfiles, err := GetAllProfiles(ctx, ndbclient)
 
+	err = GetOOBProfiles(profileMap, allProfiles, dbType)
 	if err != nil {
 		return
 	}
 
+	log.Info("Received Input Profiles = ", "Received Input Profiles", profiles)
+
+	customProfileOptions := [...]string{PROFILE_TYPE_COMPUTE, PROFILE_TYPE_SOFTWARE, PROFILE_TYPE_NETWORK, PROFILE_TYPE_DATABASE_PARAMETER}
```
manavrajvanshi, line 179: The profiles are not necessarily 'custom'. The user can also provide an OOB profile name/id as an input.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -159,18 +159,68 @@ func GetNoneTimeMachineSLA(ctx context.Context, ndbclient *ndbclient.NDBClient)
...
+	for _, profileValue := range customProfileOptions {
+		err = MatchAndSetProfile(ctx, profileValue, profiles.Compute, allProfiles, profileMap)
```
manavrajvanshi, line 181: I didn't quite understand why we're looping over the entire array and still passing only a specific type of profile value ie: `profiles.Compute` to the function.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -159,18 +159,68 @@ func GetNoneTimeMachineSLA(ctx context.Context, ndbclient *ndbclient.NDBClient)
...
+	return
+
+}
+
+/*
+This function does:
+(a) Id level matching with profiles in allProfiles
+(b) If Id level match is successful, flow proceeds to match based on versionId
+
+	When matched, the latestVersionId is overridden with the versionId as it is this attribute while dbProvisioning which is used forprofileType versionId
```
manavrajvanshi, line 196: As discussed earlier, we might not want to use the versionId. Also, adding filtration by name is one of the requirements based on the new scope of this project.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -159,18 +159,68 @@ func GetNoneTimeMachineSLA(ctx context.Context, ndbclient *ndbclient.NDBClient)
...
+	if len(matchedProfile) > 0 {
+		log.Info("Id Matched for profileType", "profileType", profileType)
+		matchedProfileByVersionId = util.Filter(matchedProfile[0].Versions, func(versions Version) bool { return versions.Id == profile.VersionId })
+		if len(matchedProfileByVersionId) > 0 {
+			log.Info("VersionId Matched for profileType", "profileType", profileType)
+			matchedProfile[0].LatestVersionId = profile.VersionId
+		}
+		profilesMap[profileType] = matchedProfile[0]
+		return
+	}
```
manavrajvanshi, line 216: As discussed earlier, we're specifically looking for a single element here, we can make use of another utility function like `findFirst` that returns the first match based on the filtration criteria rather than getting an array/slice and taking the first element.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -227,6 +292,11 @@ func GetDatabasePortByType(dbType string) int32 {
 	}
 }
 
+func isEmptyProfile(customProfile Profile) (isEmpty bool) {
+	isEmpty = customProfile == (Profile{})
+	return
+}
```
manavrajvanshi, line 298: Although this function has the correct functionality, I feel we can change the argument to say `profile` instead of `customProfile`. This function will work for any kind of profiles and we might not want add semantic meaning to the parameter name by prefixing it with 'custom'.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -159,18 +159,68 @@ func GetNoneTimeMachineSLA(ctx context.Context, ndbclient *ndbclient.NDBClient)
...
+	if isEmptyProfile(profile) {
+		log.Info("Defaulting to OOB profile population for profileType = ", "profileType", profileType)
+		return
+	}
+	var matchedProfile []ProfileResponse
+	var matchedProfileByVersionId []Version
+	matchedProfile = util.Filter(allProfiles, func(p ProfileResponse) bool { return p.Id == profile.Id && p.Type == profileType })
+	if len(matchedProfile) > 0 {
+		log.Info("Id Matched for profileType", "profileType", profileType)
+		matchedProfileByVersionId = util.Filter(matchedProfile[0].Versions, func(versions Version) bool { return versions.Id == profile.VersionId })
```
manavrajvanshi, line 209: Does this variable store a profile or a version of a profile?

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -159,18 +159,68 @@ func GetNoneTimeMachineSLA(ctx context.Context, ndbclient *ndbclient.NDBClient)
...
+		if len(matchedProfileByVersionId) > 0 {
+			log.Info("VersionId Matched for profileType", "profileType", profileType)
+			matchedProfile[0].LatestVersionId = profile.VersionId
+		}
+		profilesMap[profileType] = matchedProfile[0]
+		return
+	}
+	err = fmt.Errorf("No matching profile found for profileType = %s", profileType)
```
manavrajvanshi, line 217: This might throw linting errors, error strings should not be capitalized.

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -195,10 +245,25 @@ func GetOOBProfiles(ctx context.Context, ndbclient *ndbclient.NDBClient, dbType
...
+	case PROFILE_TYPE_SOFTWARE:
+		return profiles.Software
+	case PROFILE_TYPE_NETWORK:
+		return profiles.Network
+	case PROFILE_TYPE_DATABASE_PARAMETER:
+		return profiles.DbParam
+	default:
+		return defaultEmptyProfile
+	}
+}
```
manavrajvanshi, line 265: This function is not used in the business logic, I only see it in one of the test files. How do you feel about placing it under the test files instead?
(Since it does not add any functionality/value in the ndb_api_helpers.go file)

api/v1alpha1/ndb_api_helpers.go
manavrajvanshi, line 179: Should we move this to the constants file instead?

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -159,18 +159,68 @@ func GetNoneTimeMachineSLA(ctx context.Context, ndbclient *ndbclient.NDBClient)
...
 	}
 
+	log.Info("Received Input Profiles = ", "Received Input Profiles", profiles)
+
+	customProfileOptions := [...]string{PROFILE_TYPE_COMPUTE, PROFILE_TYPE_SOFTWARE, PROFILE_TYPE_NETWORK, PROFILE_TYPE_DATABASE_PARAMETER}
+	for _, profileValue := range customProfileOptions {
+		err = MatchAndSetProfile(ctx, profileValue, profiles.Compute, allProfiles, profileMap)
+		if err != nil {
+			return
+		}
```
manavrajvanshi, line 184: Can we also log the error here?

api/v1alpha1/ndb_api_helpers.go
manavrajvanshi, line 196: Thanks for adding a detailed comment, helps in understanding the complexities of the logic.

manavrajvanshi (changes requested):
Some comments around the code.
Also, @akshmish has started working on this issue, connecting you two!

manavrajvanshi:
Fixed by @akshmish in #97

PR #98, opened 2023-05-17
Defaulted db_user to admin in MongoDB provisioning

manavrajvanshi:
This PR addresses a bug which assigns the vm instance name to the db_user in mongodb provisioning.

This has been tested manually. 
I created a db instance of mongodb type (with service creation hack, since that is also a known bug at the time of writing this).
I also created an application pod with details about the service and credentials. (https://github.com/nutanix-scratch/manavrajvanshi/tree/main/best-app/mongodb)
I was able to interact with the application via postman when the database was provisioned.

image
image

None

**Release note**:

```release-note
None
```

### DIFF
--- a/api/v1alpha1/ndb_api_helpers.go
+++ b/api/v1alpha1/ndb_api_helpers.go
@@ -259,7 +259,7 @@ func GetActionArgumentsByDatabaseType(databaseType string) (DatabaseActionArgs, 
 	case DATABASE_TYPE_MONGODB:
 		dbTypeActionArgs = &MongodbActionArgs{}
 	default:
-		return nil, errors.New("Invalid Database Type: supported values: mysql, postgres, mongodb")
+		return nil, errors.New("invalid database type: supported values: mysql, postgres, mongodb")
 	}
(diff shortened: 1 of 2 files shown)


akshmish (approved):

api/v1alpha1/ndb_api_helpers.go
```diff
@@ -326,7 +326,7 @@ func (m *MongodbActionArgs) GetActionArguments(dbSpec DatabaseSpec) []ActionArgu
 		},
 		{
 			Name:  "db_user",
-			Value: dbSpec.Instance.DatabaseInstanceName,
+			Value: "admin",
```
akshmish, line 329: Is this change a stop-gap solution or do we plan to accept the db_user value as an input from the users?

manavrajvanshi: Stopgap solution. This will be tracked for taking in user inputs too.

shenoypritika (approved):

krunal-jhaveri (approved):

PR #107, opened 2023-06-05
Automatic backups

manavrajvanshi:
**What this PR does / why we need it**: This PR adds support for automatic backups / providing TM and SLA details while provisioning a database.

**How Has This Been Tested?**: This has been tested through UTs and Manual e2e tests.
image

image

**Release note**:

```release-note
Adds TM/SLA details
```


controller_adapters/database.go
```diff
@@ -82,3 +84,102 @@ func (d *Database) GetProfileResolvers() ndb_api.ProfileResolvers {
...
+	switch tmInfo.QuarterlySnapshots {
+	case "Jan":
+		quarterlySnapshotStartMonth = "JANUARY"
+	case "Feb":
+		quarterlySnapshotStartMonth = "FEBRUARY"
+	case "Mar":
+		quarterlySnapshotStartMonth = "MARCH"
+	default:
+		quarterlySnapshotStartMonth = ""
+	}
```
manavrajvanshi, line 147: Open Question: Should we error out if no value is provided?

ndb_api/utility_test.go
```diff
@@ -23,205 +23,13 @@ import (
...
-			Name:            "DEFAULT_OOB_MYSQL_NETWORK",
-			Type:            common.PROFILE_TYPE_NETWORK,
-			EngineType:      common.DATABASE_ENGINE_TYPE_MYSQL,
-			LatestVersionId: "v-id-6",
-			Topology:        common.TOPOLOGY_SINGLE,
-			Status:          "READY",
-			SystemProfile:   true,
-		},
-	},
-}
```
manavrajvanshi, line 224: Moved to `ndb_api/mock_responses_test.go`

test/ndb_api_helpers_test.go
```diff
@@ -573,450 +505,3 @@ func TestResolveComputeProfileById_resolvesOk(t *testing.T) {
...
-			Spec: dbSpec,
-		}}
-
-		_, err := ndb_api.GenerateProvisioningRequest(context.Background(), ndb_client, db, reqData)
-		t.Log(err)
-		if err == nil {
-			t.Errorf("GenerateProvisioningRequest should return an error when ssh key is empty")
-		}
-	}
-}
```
manavrajvanshi, line 1022: Refactored to use table test format and moved to `ndb_api/db_helpers_test.go` to be consistent with the convention set by @shenoypritika.

api/v1alpha1/database_types.go
```diff
@@ -110,6 +110,57 @@ type Instance struct {
...
 	// +optional
 	Profiles Profiles `json:"profiles"`
+	// +optional
+	// +kubebuilder:default:="{emptytminfo}"
+	// Information related to time machine that is to be associated with this database
+	TMInfo TimeMachineInfo `json:"timeMachine"`
+}
+
+// Time Machine details
+type TimeMachineInfo struct {
```
yash-ntnx, line 120: should we move this to a diff file under the same package?

controller_adapters/database.go
```diff
@@ -82,3 +84,102 @@ func (d *Database) GetProfileResolvers() ndb_api.ProfileResolvers {
...
+		slaName = common.SLA_NAME_NONE
+	}
+
+	return
+}
+
+// Returns a schedule struct for the time machine.
+func (d *Database) GetTMSchedule() (schedule ndb_api.Schedule, err error) {
+	tmInfo := d.Spec.Instance.TMInfo
+	hhmmssDaily := strings.Split(tmInfo.DailySnapshotTime, ":")
```
yash-ntnx, line 116: can we factor out the ability to parse `tmInfo.DailySnapshotTime` in hh, mm, ss into a separate fn?

controller_adapters/database.go
yash-ntnx: this is an optional input in the spec, right? what happens in practice if the user does not provide this?

controller_adapters/database.go
```diff
@@ -82,3 +84,102 @@ func (d *Database) GetProfileResolvers() ndb_api.ProfileResolvers {
...
+	if len(hhmmssDaily) != 3 {
+		err = errors.New("invalid DailySnapshotTime, use the (24-hour) hh:mm:ss format")
+		return
```
yash-ntnx, line 119: if err != nil, we should return an invalid `schedule` (nil value instead of a zero-initialized object). 
if one return value is non-nil (in this example `err`), the other should/cannot also be non-nil (`schedule` in this example). I am not sure of the implications of this while writing UTs because while mocking the response we are expecting a zero-initialized object at the moment. Not sure if we can mock/expect `nil`

controller_adapters/database.go
manavrajvanshi: AFAIK, it gets a default value on the UI.

api/v1alpha1/database_types.go
manavrajvanshi: Yes we can! But I kept it under `database_types.go` because this is closely related to the database spec. If I move this out to something like TimeMachine/TimeMachineInfo.go and sometime down the line when we introduce the TimeMachine CR, then that will create something like `timemachine_types.go`, this might create some confusion which is what I was trying to avoid. Anyway, let me know how you feel about this.

controller_adapters/database.go
manavrajvanshi: Sure, will move that out to a non-exported function.

controller_adapters/database.go
manavrajvanshi: With the current implementation, it is assumed that the client/caller always checks for the err before making use of the returned `schedule` object. But yes, I see your point. Though I cannot return a nil schedule/struct in golang, I can change the return type to be a pointer to the struct and then return a nil pointer. Sounds good?

controller_adapters/database.go
manavrajvanshi: I realized that there's a `time.Parse` function available in golang, replaced the code block with that instead.

`hhmmss, err := time.Parse(time.TimeOnly, tmInfo.DailySnapshotTime)
	if err != nil {
		err = errors.Join(err, errors.New("error converting daily snapshot time"))
		return
	}
	hh, mm, ss := hhmmss.Hour(), hhmmss.Minute(), hhmmss.Second()`

controller_adapters/database.go
manavrajvanshi: Okay, so I changed the function (and the interface definition) to return a pointer to the schedule struct. Now our test will need to check for the returned pointer value - which will not be known to us, a reference to the schedule is created within the execution scope of the function and returned and we cannot guess/provide it to the test as an expected value. 

The only check we can perform is to check if the pointer is nil or not, this purpose is being served by the nil err check exactly, making our tests adding no value. Hence I feel returning empty/zero initialized objects with a non-nil err might be the correct way. We can discuss more, let me know how you feel about this. :)

controller_adapters/database.go
manavrajvanshi: Also, if not provided in yaml input spec it gets defaulted to `Jan`.

Makefile
```diff
@@ -97,6 +97,7 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	sed -i.bak "s/'{emptytminfo}'/{}/g" config/crd/bases/*.yaml && rm config/crd/bases/*.bak
```
akshmish, line 100: I remember syncing offline about this but please do add a comment explaining "why this had to be done", so the future readers have the full context.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.18 as builder
+FROM golang:1.20 as builder
```
akshmish, line 2: I am ok with this change but let's prefer to roll out version upgrades as separate changes for better trackability.

api/v1alpha1/database_types.go
```diff
@@ -110,6 +110,57 @@ type Instance struct {
 	Type string `json:"type"`
 	// +optional
 	Profiles Profiles `json:"profiles"`
+	// +optional
+	// +kubebuilder:default:="{emptytminfo}"
```
akshmish, line 114: add a comment here as well about default:="{emptytminfo}"

api/v1alpha1/database_types.go
```diff
@@ -110,6 +110,57 @@ type Instance struct {
...
+	// Day of the month for monthly snapshot
+	MonthlySnapshotDay int `json:"monthlySnapshotDay"`
+	// +kubebuilder:default=Jan
+	// +kubebuilder:validation:Enum=Jan;Feb;Mar
+	// +optional
+	// Start month for the quarterly snapshot
+	// Jan => Jan, Apr, Jul, Oct.
+	// Feb => Feb, May, Aug, Nov.
+	// Mar => Mar, Jun, Sep, Dec.
+	QuarterlySnapshots string `json:"quarterlySnapshots"`
```
akshmish, line 163: weeklySnapshotDay
monthlySnapshotDay
quarterlySnapshots"Month"

"Month" suffix is missing

controller_adapters/database.go
```diff
@@ -82,3 +84,89 @@ func (d *Database) GetProfileResolvers() ndb_api.ProfileResolvers {
...
+	}
+	if slaName == "" {
+		slaName = common.SLA_NAME_NONE
+	}
+
+	return
+}
+
+// Returns a schedule struct for the time machine.
+func (d *Database) GetTMSchedule() (schedule ndb_api.Schedule, err error) {
```
akshmish, line 114: Please revisit the field-mappings as per the offline discussions:
eg Quarterly schedule is not applicable for NONE sla or BRASS sla. 

There are also other multiple fields that UI "greys out" (based on applicability) when you select a specific SLAs from the dropdown. 

The easiest way would be to take a look at the UI code and how it disables those input fields based on the SLA selection and translate that into your logic here.

manavrajvanshi: Cool, that is a great catch.. Thanks! :)

ndb_api/db_helpers.go
```diff
@@ -32,10 +32,18 @@ func GenerateProvisioningRequest(ctx context.Context, ndb_client *ndb_client.NDB
 	log := ctrllog.FromContext(ctx)
 	log.Info("Entered ndb_api.GenerateProvisioningRequest", "database name", database.GetDBInstanceName(), "database type", database.GetDBInstanceType())
 
+	// Fetching the TM details
+	tmName, tmDescription, slaName := database.GetTMDetails()
 	// Fetching the NONE TM SLA
```
shenoypritika, line 37: nit: Can we update the comment here

ndb_api/db_helpers_test.go
```diff
@@ -405,3 +409,287 @@ func TestMySqlProvisionRequestAppender(t *testing.T) {
...
+		if err != tc.expectedError && err.Error() != tc.expectedError.Error() {
+			t.Fatalf("expected: %v, got: %v", tc.expectedError, err)
+		}
+	}
+}
+
+// Test the error scenarios in GenerateProvisioningRequest function with different ProfileResolver errors
+// 1. Software Profile returns an error
+// 2. Compute Profile returns an error
+// 3. Network Profile returns an error
```
shenoypritika, line 477: Great job on the comments!

ndb_api/db_helpers.go
manavrajvanshi: Great catch, thanks!

controller_adapters/database.go
manavrajvanshi: Update: It only gets defaulted to `Jan` when the entire TMInfo object is not provided in the input spec. otherwise this field remains empty. Hence I feel we should default it to "Jan" explicitly here.

controller_adapters/database.go
manavrajvanshi: So I looked into the UI code as suggested. Although the fields are greyed out for certain SLA selections, the API request payload does not change. This means that the API is ignoring the non-mandatory inputs for these SLAs. 

To confirm, I provisioned a database using a REST API call via postman with a NONE SLA and non-empty values for the fields in the schedule. The provisioning went through successfully. 

What should we be doing in this scenario?

akshmish: By greying out the fields based on the SLA selected, the UI makes it explicit to the user that those "fields are not required." How do we similarly make it explicit to the user using the NDB operator. Webhooks is one option here. So those validations may need to be implemented once the webhooks PR is merged. That way, if and when the validation is added in the APIs in the future, we will not need to change the operator code.  

For now, you may choose to send the default values for those keys in the payload since it has been verified that sending default values for those non-applicable does not adversely impact the provisioning flow. Also the same is being done by the UI code too.

api/v1alpha1/database_types.go
manavrajvanshi: I added it, but then `make generate manifests` added the comment in the description of the CRD yaml, had to remove it ultimately. Nonetheless, I have provided information about this in the makefile comment and have added the relevant GitHub issue too.

api/v1alpha1/database_types.go
yash-ntnx: talked offline - we will rename this struct and keep it at the same location

controller_adapters/database.go
yash-ntnx: talked offline - returning a nil pointer to the struct will cause the operator to panic in case someone tries to reference it. we will leave the behavior as-is for now.

Makefile
```diff
@@ -94,9 +94,16 @@ help: ## Display this help.
 
 ##@ Development
 
+## Explanation of the statement: sed -i.bak "s/'{emptytminfo}'/{}/g" config/crd/bases/*.yaml && rm config/crd/bases/*.bak
+## This is an open issue on operator-sdk: https://github.com/operator-framework/operator-sdk/issues/6346
+## TimeMachineInfo is an optional nested struct under spec section of the database CRD. However, it does not get an empty struct value '{}'
```
yash-ntnx, line 99: change TimeMachineInfo to DBTimeMachineInfo.

yash-ntnx (approved):

Makefile
manavrajvanshi: Whoops, missed that! 
Thanks :)

shenoypritika (approved):
Minor comments. LGTM!

controller_adapters/database.go
```diff
@@ -82,3 +85,92 @@ func (d *Database) GetProfileResolvers() ndb_api.ProfileResolvers {
...
+	hhmmss, err := time.Parse(time.TimeOnly, tmInfo.DailySnapshotTime)
+	if err != nil {
+		err = errors.Join(err, errors.New("error converting daily snapshot time"))
+		return
+	}
+	hh, mm, ss := hhmmss.Hour(), hhmmss.Minute(), hhmmss.Second()
+
+	var quarterlySnapshotStartMonth string
+	switch tmInfo.QuarterlySnapshotMonth {
+	case "Jan":
```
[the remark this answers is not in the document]
manavrajvanshi: Sure, thanks for the idea! Will make it less verbose (:

PR #108, opened 2023-06-05
Adding go-validator for validating the user-defined structs

akshmish:
Adds a standard validator library for validating the user-defined structs (that cannot be validated using Kubebuilder's annotation based validators)

go-validator: https://github.com/go-playground/validator 

WIP

**Release note**:
2. If no release note is required, just write "NONE".

```release-note
NONE
```


controllers/database_reconciler_helpers.go
```diff
@@ -222,12 +222,18 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
 
 		databaseAdapter := &controller_adapters.Database{Database: *database}
 		generatedReq, err := ndb_api.GenerateProvisioningRequest(ctx, ndbClient, databaseAdapter, reqData)
-		// generatedReq, err := ndb_api.GenerateProvisioningRequestt(ctx, ndbClient, database.Spec, reqData)
+
 		if err != nil {
-			log.Error(err, "Could not generate provisioning request, requeuing.")
+			log.Error(err, "Could not generate provisioning request, requeuing...")
 			return r.requeueOnErr(err)
 		}
```
manavrajvanshi, line 229: [Non Blocking/General Discussion] - Since you're already working on this file/improvement: How do you feel about moving the request generation into the ProvisionDatabase call itself?

ndb_api/common_helpers.go
```diff
@@ -45,3 +46,8 @@ func GetDatabasePortByType(dbType string) int32 {
 		return -1
 	}
 }
+
+func ValidateDatabaseProvisionRequest(req DatabaseProvisionRequest) error {
+	err := validator.New().Struct(req)
+	return err
+}
```
manavrajvanshi, line 53: We might not need this at all when we move the validation into ndb_api.ProvisionDatabase function..

manavrajvanshi:
Closing since the validations will go in the webhook PR.

PR #111, opened 2023-06-09
Support mssql provisioning

shenoypritika:
- Add support for MSSQL provisioning. 

- Provisioning and service creation works fine:

image

image

- App connects to the k8s service 

image


akshmish:
Can you please populate the PR template?

ndb_api/db_helpers.go
```diff
@@ -142,4 +109,37 @@ func GenerateDeprovisionDatabaseRequest() (req *DatabaseDeprovisionRequest) {
...
 	return
+
+}
+
+func validateReqData(ctx context.Context, databaseInstanceType string, reqData map[string]interface{}) (err error) {
+	log := ctrllog.FromContext(ctx)
+	dbPassword, ok := reqData[common.NDB_PARAM_PASSWORD].(string)
+	// Type Assertion for dbPassword
+	if !ok || dbPassword == "" {
+		err = errors.New("invalid database password")
```
akshmish, line 120: "No password provided" error message makes more sense?

akshmish, line 120: Curious when the "type assertion" may fail for password? If we do not get into that case, the error message can simply be updated to "empty database password"

ndb_api/db_request_types.go
```diff
@@ -29,12 +29,13 @@ type DatabaseProvisionRequest struct {
 	CreateDbServer           bool             `json:"createDbserver"`
 	NodeCount                int              `json:"nodeCount"`
 	NxClusterId              string           `json:"nxClusterId"`
-	SSHPublicKey             string           `json:"sshPublicKey"`
+	SSHPublicKey             string           `json:"sshPublicKey,omitempty"`
```
akshmish, line 32: Sorry can you educate me on this one? Doesn't "omitempty" cause the generated CRD to mark this field as optional? That is not an optional field in our APIs right?

ndb_api/db_request_types.go
```diff
@@ -29,12 +29,13 @@ type DatabaseProvisionRequest struct {
...
 	Clustered                bool             `json:"clustered"`
 	AutoTuneStagingDrive     bool             `json:"autoTuneStagingDrive"`
 	TimeMachineInfo          TimeMachineInfo  `json:"timeMachineInfo"`
 	ActionArguments          []ActionArgument `json:"actionArguments"`
 	Nodes                    []Node           `json:"nodes"`
+	DatabaseName             string           `json:"databaseName,omitempty"`
```
akshmish, line 38: same here

common/constants.go
```diff
@@ -36,6 +36,7 @@ const (
 	DATABASE_TYPE_GENERIC   = "generic"
 	DATABASE_TYPE_MONGODB   = "mongodb"
 	DATABASE_TYPE_MYSQL     = "mysql"
+	DATABASE_TYPE_MSSQL     = "mssql"
 	DATABASE_TYPE_ORACLE    = "oracle"
```
manavrajvanshi, line 40: Please maintain the alphabetical order.

ndb_api/db_helpers.go
```diff
@@ -142,4 +109,37 @@ func GenerateDeprovisionDatabaseRequest() (req *DatabaseDeprovisionRequest) {
...
+			errStatement = "Empty database password"
+		}
+		log.Error(err, errStatement)
+	}
+
+	// Type Assertion for SSHKey
+	if databaseInstanceType != common.DATABASE_TYPE_MSSQL {
+		SSHPublicKey, ok := reqData[common.NDB_PARAM_SSH_PUBLIC_KEY].(string)
+		if !ok || SSHPublicKey == "" {
+			err := errors.New("invalid ssh public key")
```
manavrajvanshi, line 134: Change `:=` to `=`

ndb_api/db_helpers.go
```diff
@@ -46,32 +46,7 @@ func GenerateProvisioningRequest(ctx context.Context, ndb_client *ndb_client.NDB
...
-		err = errors.New("invalid ssh public key")
-		var errStatement string
-		if !ok {
-			errStatement = "Type assertion failed for SSHPublicKey. Expected a string value"
-		} else {
-			errStatement = "Empty SSHPublicKey"
-		}
-		log.Error(err, errStatement)
-	}
+	validateReqData(ctx, database.GetDBInstanceType(), reqData)
```
manavrajvanshi, line 49: Please check for the error.

ndb_api/request_appender.go
```diff
@@ -0,0 +1,221 @@
...
+	"errors"
+
+	"github.com/nutanix-cloud-native/ndb-operator/common"
+)
+
+type DBProvisionRequestAppender interface {
+	appendRequest(req *DatabaseProvisionRequest, database DatabaseInterface, reqData map[string]interface{}) *DatabaseProvisionRequest
+}
+
+type MSSqlProvisionRequestAppender struct{}
```
manavrajvanshi, line 13: Can we use Mssql or MsSQL or MSSQL instead?

ndb_api/request_appender.go
```diff
@@ -0,0 +1,221 @@
...
+			Name:  "server_collation",
+			Value: "SQL_Latin1_General_CP1_CI_AS",
+		},
+		{
+			Name:  "database_collation",
+			Value: "SQL_Latin1_General_CP1_CI_AS",
+		},
+		{
+			Name:  "dbParameterProfileIdInstance",
+			Value: "af9729f6-ef72-43ac-9e02-633ad13c8d51",
```
manavrajvanshi, line 73: Please check the hardcoded value.

ndb_api/request_appender.go
```diff
@@ -0,0 +1,221 @@
...
+
+	req.ActionArguments = append(req.ActionArguments, actionArgs...)
+
+	return req
+}
+
+func (a *MongoDbProvisionRequestAppender) appendRequest(req *DatabaseProvisionRequest, database DatabaseInterface, reqData map[string]interface{}) *DatabaseProvisionRequest {
+	dbPassword := reqData[common.NDB_PARAM_PASSWORD].(string)
+	databaseNames := database.GetDBInstanceDatabaseNames()
+	SSHPublicKey := reqData[common.NDB_PARAM_SSH_PUBLIC_KEY].(string)
```
manavrajvanshi, line 90: Can/should we avoid using these temp variables if they are not being used in other places?

ndb_api/request_appender.go
manavrajvanshi, line 1: I think we should move the contents of this file into the `ndb_api.db_helpers.go`, because:
These functions will only be used by db.go/db_helpers.go for generating a provisioning request.

ndb_api/db_helpers.go
```diff
@@ -109,22 +83,15 @@ func GenerateProvisioningRequest(ctx context.Context, ndb_client *ndb_client.NDB
...
 			{
 				Name:  "database_size",
 				Value: strconv.Itoa(database.GetDBInstanceSize()),
 			},
 		},
 	}
-	// Setting action arguments based on database type
-	requestBody.ActionArguments = append(requestBody.ActionArguments, database.GetDBInstanceActionArguments()...)
+	// Appending request body based on database type
+	reqAppender, err := GetDbProvRequestAppender(database.GetDBInstanceType())
```
manavrajvanshi, line 93: Can we call it requestAppender instead of reqAppender since the variable in the next line says `requestBody`. Just trying to be consistent in the same function.

manavrajvanshi (changes requested):
Added few comments.

ndb_api/db_request_types.go
shenoypritika: The SSHKey field is not required for MSSQL Databases. That is why I marked it as omitempty. There is engine specific validation for this field in the controller that would error out if SSHKey is not present for other engines

ndb_api/db_request_types.go
shenoypritika: This field is required for MSSQL Databases only. We could discuss offline how we can validate these engine specific fields.

ndb_api/db_helpers.go
shenoypritika: Oops missed that!  Added...

Thanks

ndb_api/request_appender.go
shenoypritika: Thanks for pointing this out. I have made the changes for resolving Database Parameter Instance Profile

ndb_api/request_appender.go
shenoypritika: That makes sense. Done!

ndb_api/db_helpers_test.go
```diff
@@ -0,0 +1,406 @@
...
+		},
+		{ //Throw error when password is empty
+			databaseType: common.DATABASE_TYPE_POSTGRES,
+			reqData: reqData{
+				common.NDB_PARAM_PASSWORD:       "",
+				common.NDB_PARAM_SSH_PUBLIC_KEY: TEST_SSHKEY,
+			},
+			expected: errorInvalidPassword,
+		},
+		{ // Throw error if datbase is not MSSQL and SSHKey is empty
```
yash-ntnx, line 50: nit: typo in 'datbase'

yash-ntnx (approved):

ndb_api/db_helpers.go
```diff
@@ -142,4 +115,254 @@ func GenerateDeprovisionDatabaseRequest() (req *DatabaseDeprovisionRequest) {
...
+
+type MSSQLProvisionRequestAppender struct{}
+
+type MongoDbProvisionRequestAppender struct{}
+
+type PostgresProvisionRequestAppender struct{}
+
+type MySqlProvisionRequestAppender struct{}
+
+func (a *MSSQLProvisionRequestAppender) appendRequest(req *DatabaseProvisionRequest, database DatabaseInterface, reqData map[string]interface{}) *DatabaseProvisionRequest {
```
manavrajvanshi, line 168: Should we also return an error from the `appendRequest` function?

ndb_api/interface_mock.go
manavrajvanshi, line 1: Can you also make sure that this file is excluded from being included in the build process?
Might require changes in the dockerfile or by using some markers.

ndb_api/profile_helpers.go
```diff
@@ -52,6 +52,7 @@ func ResolveProfiles(ctx context.Context, ndb_client *ndb_client.NDBClient, data
 	softwareProfileResolver := profileResolvers[common.PROFILE_TYPE_SOFTWARE]
 	networkProfileResolver := profileResolvers[common.PROFILE_TYPE_NETWORK]
 	dbParamProfileResolver := profileResolvers[common.PROFILE_TYPE_DATABASE_PARAMETER]
+	dbParamInstanceProfileResolver := profileResolvers[common.PROFILE_TYPE_DATABASE_PARAMETER]
```
manavrajvanshi, line 55: Should this be PROFILE_TYPE_DATABASE_PARAMETER_INSTANCE?

ndb_api/profile_helpers.go
```diff
@@ -117,3 +129,7 @@ var NetworkOOBProfileResolver = func(p ProfileResponse) bool {
 var DbParamOOBProfileResolver = func(p ProfileResponse) bool {
 	return p.SystemProfile && p.Type == common.PROFILE_TYPE_DATABASE_PARAMETER
 }
+
+var DbParamInstanceOOBProfileResolver = func(p ProfileResponse) bool {
+	return p.SystemProfile && p.Type == common.PROFILE_TYPE_DATABASE_PARAMETER && p.Topology == common.TOPOLOGY_INSTANCE
+}
```
manavrajvanshi, line 135: Can we also add a comment here that explains your findings regarding the overloaded use of "topology" for db param profiles? Might save/help us in the future.

manavrajvanshi (changes requested):

ndb_api/db_helpers.go
akshmish: Discussed this offline. Since "map[string]interface{}" can accept any datatype (string, int, map, object etc) as "value", we need to assert.

ndb_api/db_helpers.go
shenoypritika: Yes if we add the type assertion validations, we would need to

ndb_api/db_helpers.go
```diff
@@ -142,4 +115,254 @@ func GenerateDeprovisionDatabaseRequest() (req *DatabaseDeprovisionRequest) {
...
+	}
+
+	// Type Assertion for SSHKey
+	if databaseInstanceType != common.DATABASE_TYPE_MSSQL {
+		SSHPublicKey, ok := reqData[common.NDB_PARAM_SSH_PUBLIC_KEY].(string)
+		if !ok || SSHPublicKey == "" {
+			err = errors.New("invalid ssh public key")
+			var errStatement string
+			if !ok {
+				errStatement = "Type assertion failed for SSHPublicKey. Expected a string value"
```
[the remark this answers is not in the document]
shenoypritika: Will address this in the next PR

ndb_api/interface_mock.go
shenoypritika: As suggested renamed it to interface_mock_test

ndb_api/profile_helpers.go
```diff
@@ -86,15 +87,26 @@ func ResolveProfiles(ctx context.Context, ndb_client *ndb_client.NDBClient, data
...
-		log.Error(err, "DbParam Profile could not be resolved", "Input Profile", networkProfileResolver)
+		log.Error(err, "DbParam Profile could not be resolved", "Input Profile", dbParamProfileResolver)
 		return nil, err
 	}
 
+	// DB Param Instance Profile
+	dbParamInstance, err := dbParamInstanceProfileResolver.Resolve(ctx, dbEngineSpecific, DbParamOOBProfileResolver)
+	if err != nil {
+		// Database Parameter Instance profile is required only for sql server
+		if databaseType == common.DATABASE_TYPE_SQLSERVER {
```
manavrajvanshi, line 98: possible duplication: MSSQL/SQLSERVER, please check.

ndb_api/profile_helpers.go
```diff
@@ -86,15 +87,26 @@ func ResolveProfiles(ctx context.Context, ndb_client *ndb_client.NDBClient, data
...
+			log.Error(err, "Db Param Instance Profile could not be resolved", "Input Profile", dbParamInstanceProfileResolver)
+			return nil, err
+		}
```
manavrajvanshi, line 101: Also, I feel we should add an `else` block where we set the `err` to `nil`. 
Think of the scenario when we're provisioning a postgres db for example. We will not find a dbParamInstance profile. The resolver will return an error and we'll skip printing the error since it is not a MSSQL database. However, the err variable still holds an error and when the ResolveProfiles function returns, it will return the error in the `err` variable. 
WDYT?

manavrajvanshi (changes requested):
Thanks for introducing table tests and mocks. This will reduce the clutter and enhance readability. Left some minor comments, let's discuss those if needed. Also, please verify the deployment/connectivity etc. Rest LGTM!

ndb_api/profile_helpers.go
shenoypritika: Makes sense! Done

ndb_api/profile_helpers.go
shenoypritika: Fixed, thanks for catching this

manavrajvanshi (approved):
LGTM, Thanks!  💯

PR #116, opened 2023-06-15
Adds automated tests framework

manavrajvanshi:
**What this PR does / why we need it**: The PR adds a framework for adding automation tests.

**Special notes for your reviewer**
Please review the setups and teardown methods with a greater detail.

**Release note**:

```release-note
None
```


Makefile
```diff
@@ -112,7 +112,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test `go list ./... | grep -v "/automation"` -coverprofile cover.out
```
manavrajvanshi, line 115: Added to exclude running the automation tests while running unit tests.

automation/helpers.go
```diff
@@ -0,0 +1,92 @@
...
+		log.Println(err)
+	}
+	err = json.Unmarshal(jsonData, secret)
+	if err != nil {
+		log.Println(err)
+	}
+	return
+}
+
+func ConvertToDatabase(data []byte) (database *ndbv1alpha1.Database, err error) {
```
yash-ntnx, line 37: all of these utility functions essentially unmarshal a byte stream into an object of interest (secret, pod, database, etc).
can we just have 1 func that does all this? it takes "type" as an argument and maybe we can do some magic with the reflect package. can you pls see if that is possible?

automation/setup.go
```diff
@@ -0,0 +1,196 @@
...
+	clientsetv1alpha1 "github.com/nutanix-cloud-native/ndb-operator/automation/clientset/v1alpha1"
+	"github.com/nutanix-cloud-native/ndb-operator/common"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func setup(dbSecret, ndbSecret *corev1.Secret, database *ndbv1alpha1.Database, appSvc *corev1.Service, appPod *corev1.Pod, clientset *kubernetes.Clientset, v1alpha1ClientSet *clientsetv1alpha1.V1alpha1Client, t *testing.T) (err error) {
+
+	ns := "default"
```
shenoypritika, line 21: Could we add this to a constant file?

automation/helpers.go
```diff
@@ -0,0 +1,92 @@
...
+	secret = &corev1.Secret{}
+	jsonData, err := yaml.YAMLToJSON(data)
+	if err != nil {
+		log.Println(err)
+	}
+	err = json.Unmarshal(jsonData, secret)
+	if err != nil {
+		log.Println(err)
+	}
+	return
```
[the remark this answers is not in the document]
manavrajvanshi: Oh, that is because of the named return values defined in the function signature and the naked return.

automation/setup.go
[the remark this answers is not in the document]
manavrajvanshi: I don't think there will be any conflict between the function names and the file name. But we can append the test_ prefix to function names.

shenoypritika (approved):

PR #124, opened 2023-07-06
Add error handling in GenerateProvisioningRequest

shenoypritika:
- The error returned by GetDbProvRequestAppender was not handled. This could possibly lead to a NPE if it returns nil. This PR handles the error.

- Provisioning completed successfully

image

image


ndb_api/db_helpers_test.go
```diff
@@ -693,3 +693,62 @@ func TestGenerateProvisioningRequest_AgainstDifferentReqData(t *testing.T) {
...
+
+	// Test
+	_, err := GenerateProvisioningRequest(context.Background(), ndb_client, &mockDatabase, reqData)
+	expectedError := errors.New("invalid database type: supported values: mssql, mysql, postgres, mongodb")
+
+	// Assert
+	if err != expectedError && err.Error() != expectedError.Error() {
+		t.Fatalf("expected: %v, got: %v", expectedError, err)
+	}
+}
```
manavrajvanshi, line 754: Can we merge this test with the reqData test as these are similar, also you can change the test name to something more generic after merging.

manavrajvanshi (changes requested):
Please refactor the UT to fit into the existing test. Thanks!

ndb_api/db_helpers_test.go
shenoypritika: Done!

config/crd/bases/ndb.nutanix.com_databases.yaml
```diff
@@ -119,6 +119,7 @@ spec:
                       associated with this database
                     properties:
                       dailySnapshotTime:
+                        default: "12:34:56"
```
yash-ntnx, line 122: it is better to have a default snapshot time in the middle of the night when a small DB downtime is easier to tolerate.

shenoypritika: Sure! Changed it!

manavrajvanshi (approved):
LGTM, Thanks!

yash-ntnx (approved):

PR #129, opened 2023-07-13
Test for Web-hooks and converting the top-level Instance Spec fields to pointers

akshmish:

Converts the structs input fields into Pointers.
Add Functional Tests for the Web-hooks

Test Spec:
https://docs.google.com/document/d/1bxNGsA3JU9bktCfjEbYoq51kQLqi9QTGiejK1D0THCc/edit


Makefile
```diff
@@ -37,7 +37,7 @@ PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
 # nutanix.com/ndb-operator-bundle:$VERSION and nutanix.com/ndb-operator-catalog:$VERSION.
-IMAGE_TAG_BASE ?= ghcr.io/nutanix-cloud-native/ndb-operator/controller
```
akshmish, line 40: Leaving a note here for myself to delete this before merging.

api/v1alpha1/database_webhook.go
```diff
@@ -58,41 +68,46 @@ func instanceSpecDefaulterForCreate(r *Database) {
...
-	if tmInfo.MonthlySnapshotDay == 0 {
-		tmInfo.MonthlySnapshotDay = 15
+	if r.Spec.Instance.TMInfo.MonthlySnapshotDay == 0 {
+		r.Spec.Instance.TMInfo.MonthlySnapshotDay = 15
 	}
 
-	if tmInfo.QuarterlySnapshotMonth == "" {
-		tmInfo.QuarterlySnapshotMonth = "Jan"
+	if r.Spec.Instance.TMInfo.QuarterlySnapshotMonth == "" {
+		r.Spec.Instance.TMInfo.QuarterlySnapshotMonth = "Jan"
```
manavrajvanshi, line 110: Please change all the occurrances of `r.Spec.Instance.TMInfo` to a single variable. Example: 
`instance := r.Spec.Instance.TMInfo`. This will help readbability and maintainability of the code.tmInfo

api/v1alpha1/database_webhook.go
```diff
@@ -58,41 +68,46 @@ func instanceSpecDefaulterForCreate(r *Database) {
 
 	// time machine defaulter logic
 
-	tmInfo := r.Spec.Instance.TMInfo
-	if tmInfo.Name == "" {
-		tmInfo.Name = r.Spec.Instance.DatabaseInstanceName + "_TM"
+	// initialize TM block if it's not provided by the user
+	if r.Spec.Instance.TMInfo == nil {
+		databaselog.Info("Initialzing to empty...", "tmInfo", r.Spec.Instance.TMInfo)
```
manavrajvanshi, line 73: Can we just log a single statement like Initializing TMInfo with default values? 
The current log would be something like this: `Initializing to empty... {CONTEXT DATA, "tmInfo":"{}"}`, this will not be very helpful in understanding what is being initialized with default values. Also, the empty value `{}` doesn't add any value to the log, what do you think?

api/v1alpha1/database_webhook.go
```diff
@@ -44,12 +46,20 @@ var _ webhook.Defaulter = &Database{}
...
-		r.Spec.Instance.TimeZone = "UTC"
+	if r.Spec.Instance.TimeZone == nil || *r.Spec.Instance.TimeZone == "" {
+		utc := common.TIMEZONE_UTC
+		r.Spec.Instance.TimeZone = &utc
+	}
+
+	// initialize Profiles block if it's not provided by the user
+
+	if r.Spec.Instance.Profiles == nil {
+		databaselog.Info("Initialzing to empty...", "profiles", r.Spec.Instance.Profiles)
```
manavrajvanshi, line 61: Same as TMInfo log..

api/v1alpha1/database_webhook.go
```diff
@@ -135,25 +150,28 @@ func ndbServerSpecValidatorForCreate(r *Database, allErrs field.ErrorList, ndbPa
 func instanceSpecValidatorForCreate(r *Database, allErrs field.ErrorList, instancePath *field.Path) field.ErrorList {
 	databaselog.Info("Entering instanceSpecValidatorForCreate...")
 
-	if r.Spec.Instance.DatabaseInstanceName == "" {
+	// need to ass rert using a regex
```
manavrajvanshi, line 153: This comment can be fixed. Profanity, lol.

api/v1alpha1/database_webhook.go
```diff
@@ -135,25 +150,28 @@ func ndbServerSpecValidatorForCreate(r *Database, allErrs field.ErrorList, ndbPa
...
+			fmt.Sprintf("A valid database type must be specified. Valid values are: %s", reflect.ValueOf(api.AllowedDatabaseTypes).MapKeys()),
+		))
 	}
 
-	if _, isPresent := api.ClosedSourceDatabaseTypes[r.Spec.Instance.Type]; isPresent {
-		if r.Spec.Instance.Profiles == (Profiles{}) || r.Spec.Instance.Profiles.Software == (Profile{}) {
-			allErrs = append(allErrs, field.Invalid(instancePath.Child("profiles"), r.Spec.Instance.Profiles, "Software Profile must be provided for the closed-source database engines"))
+	if _, isPresent := api.ClosedSourceDatabaseTypes[*r.Spec.Instance.Type]; isPresent {
+		if r.Spec.Instance.Profiles == &(Profiles{}) || r.Spec.Instance.Profiles.Software == (Profile{}) {
+			allErrs = append(allErrs, field.Invalid(instancePath.Child("profiles").Child("software"), r.Spec.Instance.Profiles.Software, "Software Profile must be provided for the closed-source database engines"))
```
manavrajvanshi, line 174: Extract r.Spec.Instance to instance

common/constants.go
```diff
@@ -55,6 +55,8 @@ const (
 
 	PROFILE_STATUS_READY = "READY"
 
+	TIMEZONE_UTC = "UTC"
```
manavrajvanshi, line 58: Nit: Move this before TOPOLOGY constants and after SLA.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: ghcr.io/nutanix-cloud-native/ndb-operator/controller
+  newName: 089786/ndb-operator
```
manavrajvanshi, line 15: Please revert before merging.

test/ndb_api_helpers_test.go
```diff
@@ -29,6 +29,13 @@ import (
 )
 
 func GetProfileResolvers(d v1alpha1.Database) ndb_api.ProfileResolvers {
+
+	// Webhook defaults tje Profiles to empty struct, but while running the UTs,
```
manavrajvanshi, line 33: Fix typo

manavrajvanshi (changes requested):

api/v1alpha1/database_types.go
```diff
@@ -84,25 +84,24 @@ type NDB struct {
...
+	Type     *string `json:"type"`
 	// +optional
-	Profiles Profiles `json:"profiles"`
+	Profiles *Profiles `json:"profiles"`
 	// +optional
-	// +kubebuilder:default:="{emptytminfo}"
 	// Information related to time machine that is to be associated with this database
-	TMInfo DBTimeMachineInfo `json:"timeMachine"`
+	TMInfo *DBTimeMachineInfo `json:"timeMachine"`
 }
```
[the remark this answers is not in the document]
akshmish: As discussed in the Friday sync-up call, we just need to mark the top-level fields right now. This is because depending upon clone/new provisioning, a certain set of inputs are needed or not needed, but I am not immediately be able to see where a partial set is needed. If we come across such a case, I will convert those into pointers too.

As of now, trying not to over-optimize in anticipation.

api/v1alpha1/database_webhook.go
akshmish: Makes sense, I can simply log "Initialzing empty Profiles (or TMInfo)". Not using the "default" word since we are yet to populate the default values in the struct at that point of time. We will see only zero-values for the attributes at that point.

api/v1alpha1/database_webhook.go
akshmish: Refactored for the method to accept a pointer to an Instance object as this method is only working on Instance validations.

api/v1alpha1/database_webhook.go
akshmish: Refactored the method to accept a pointer to instance object since it is only defaulting the fields in the instance. TMInfo can be read from the instance field as needed.

api/v1alpha1/database_webhook.go
```diff
@@ -109,82 +124,91 @@ func (r *Database) Default() {
...
-		allErrs = append(allErrs, field.Invalid(tmPath.Child("monthlySnapshotDay"), tmInfo.MonthlySnapshotDay, "Monthly snapshot day value must be between 1 and 28"))
+	if instance.TMInfo.MonthlySnapshotDay < 1 || instance.TMInfo.MonthlySnapshotDay > 28 {
+		allErrs = append(allErrs, field.Invalid(tmPath.Child("monthlySnapshotDay"), instance.TMInfo.MonthlySnapshotDay, "Monthly snapshot day value must be between 1 and 28"))
 	}
 
-	if _, isPresent := api.AllowedQuarterlySnapshotMonths[tmInfo.QuarterlySnapshotMonth]; !isPresent {
-		allErrs = append(allErrs, field.Invalid(tmPath.Child("quarterlySnapshotMonth"), tmInfo.QuarterlySnapshotMonth, "Quarterly snapshot month must be one of {Jan, Feb, Mar}"))
+	if _, isPresent := api.AllowedQuarterlySnapshotMonths[instance.TMInfo.QuarterlySnapshotMonth]; !isPresent {
+		allErrs = append(allErrs, field.Invalid(tmPath.Child("quarterlySnapshotMonth"), instance.TMInfo.QuarterlySnapshotMonth,
+			fmt.Sprintf("Quarterly snapshot month must be specified. Valid values are: %s", reflect.ValueOf(api.AllowedQuarterlySnapshotMonths).MapKeys()),
```
shenoypritika, line 210: Can we store the error message a separate variable and use that in allErrs, for better readability.

api/v1alpha1/database_webhook.go
```diff
@@ -42,65 +44,78 @@ func (r *Database) SetupWebhookWithManager(mgr ctrl.Manager) error {
 
 var _ webhook.Defaulter = &Database{}
 
-func instanceSpecDefaulterForCreate(r *Database) {
+func instanceSpecDefaulterForCreate(instance *Instance) {
```
shenoypritika, line 47: Can we rename this to databaseInstance ?

shenoypritika (changes requested):

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -134,13 +134,50 @@ func (in *DatabaseStatus) DeepCopy() *DatabaseStatus {
...
+		in, out := &in.DatabaseInstanceName, &out.DatabaseInstanceName
+		*out = new(string)
+		**out = **in
+	}
 	if in.DatabaseNames != nil {
 		in, out := &in.DatabaseNames, &out.DatabaseNames
-		*out = make([]string, len(*in))
-		copy(*out, *in)
+		*out = new([]string)
+		if **in != nil {
```
yash-ntnx, line 145: these 4-5 lines of code are not very intuitive. possible to use diff variable names? or explain what are we doing here?

api/v1alpha1/database_webhook.go
```diff
@@ -42,65 +44,78 @@ func (r *Database) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
 	}
 
-	if tmInfo.QuarterlySnapshotMonth == "" {
-		tmInfo.QuarterlySnapshotMonth = "Jan"
+	if instance.TMInfo.MonthlySnapshotDay == 0 {
+		instance.TMInfo.MonthlySnapshotDay = 15
+	}
+
+	if instance.TMInfo.QuarterlySnapshotMonth == "" {
+		instance.TMInfo.QuarterlySnapshotMonth = "Jan"
```
manavrajvanshi, line 110: How do you feel about extracting instance.TMInfo into a variable and using that instead?

api/v1alpha1/database_webhook.go
```diff
@@ -109,82 +124,91 @@ func (r *Database) Default() {
...
-	if r.Spec.Instance.Size < 10 {
-		allErrs = append(allErrs, field.Invalid(instancePath.Child("size"), r.Spec.Instance.Size, "Initial Database size must be 10 GBs or more"))
+	if instance.Size == nil || *instance.Size < 10 {
+		allErrs = append(allErrs, field.Invalid(instancePath.Child("size"), instance.Size, "Initial Database size must be specified with a value 10 GBs or more"))
 	}
 
-	if r.Spec.Instance.CredentialSecret == "" {
-		allErrs = append(allErrs, field.Invalid(instancePath.Child("credentialSecret"), r.Spec.Instance.CredentialSecret, "CredentialSecret must be provided"))
+	if instance.CredentialSecret == nil || *instance.CredentialSecret == "" {
+		allErrs = append(allErrs, field.Invalid(instancePath.Child("credentialSecret"), instance.CredentialSecret, "CredentialSecret must be provided in the Instance Spec"))
```
manavrajvanshi, line 165: If we keep the primitive types to non-pointer types then we wouldn't need to double check for nil and zero values. The zero values for these types can be validated. 
Do you know if there's any specific need to know if the user did not explicitly provide any of these?

manavrajvanshi (changes requested):

api/v1alpha1/database_webhook.go
```diff
@@ -109,82 +124,91 @@ func (r *Database) Default() {
...
+	databaselog.Info("Logging the Instance details inside validator method", "instance", instance)
+
+	// need to assert using a regex
+	if instance.DatabaseInstanceName == nil || *instance.DatabaseInstanceName == "" {
+		allErrs = append(allErrs, field.Invalid(instancePath.Child("databaseInstanceName"), instance.DatabaseInstanceName, "A unique Database Instance Name must be specified"))
 	}
 
-	if r.Spec.Instance.Size < 10 {
-		allErrs = append(allErrs, field.Invalid(instancePath.Child("size"), r.Spec.Instance.Size, "Initial Database size must be 10 GBs or more"))
+	if instance.Size == nil || *instance.Size < 10 {
```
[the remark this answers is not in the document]
akshmish: UI doesn't allow less than 10 GBs.

api/v1alpha1/database_webhook.go
akshmish: I think this is better. TMInfo is a inner attribute of instance spec so I want to keep that "context" everywhere it is used.

api/v1alpha1/database_webhook.go
akshmish: This is picked from "r.Spec.Instance" field in the spec, so for consistency purpose, "Instance" is a better option.

api/v1alpha1/database_webhook.go
akshmish: I think appending in-place is short and concise. Adding to a different variable and then appending that to the allErrs will make the code verbose, without much value-add.  So I guess we can keep it the same way.

api/v1alpha1/database_webhook.go
```diff
@@ -42,65 +44,78 @@ func (r *Database) SetupWebhookWithManager(mgr ctrl.Manager) error {
 
 var _ webhook.Defaulter = &Database{}
 
-func instanceSpecDefaulterForCreate(r *Database) {
+func instanceSpecDefaulterForCreate(instance *Instance) {
 
-	if len(r.Spec.Instance.DatabaseNames) == 0 {
-		r.Spec.Instance.DatabaseNames = api.DefaultDatabaseNames
+	if instance.DatabaseNames == nil || len(*instance.DatabaseNames) == 0 {
```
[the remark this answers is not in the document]
akshmish: Good question, I was trying to cover for when someone providing the databaseNames key but not adding any names in it. That should create an empty array.

Makefile
```diff
@@ -37,7 +37,7 @@ PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
...
+IMAGE_TAG_BASE ?= 089786/ndb-operator
```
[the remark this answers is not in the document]
akshmish: Done.

config/manager/kustomization.yaml
akshmish: done

api/v1alpha1/database_webhook.go
shenoypritika: I agree its more lines of code. But I feel it is more readable. I found it a little confusing to read at first. But sure however you are comfortbale

api/v1alpha1/database_webhook.go
akshmish: Since I am already going through some additional changes, I think I can also take care of this comment along with it. I am reverting it for some of the fields where it is not really necessary to convert to the pointers. That should help the code become more readable. Cc @yash-ntnx @shenoypritika @mshaaeldin

api/v1alpha1/webhook_suite_test.go
```diff
@@ -126,7 +129,338 @@ var _ = BeforeSuite(func() {
...
+		Expect(err).To(HaveOccurred())
+
+		// Extract the error message from the error object
+		errMsg := err.(*errors.StatusError).ErrStatus.Message
+		Expect(errMsg).To(ContainSubstring("A valid Database Instance Name must be specified"))
+
+	})
+})
+
+var _ = Describe("Webhooks: CredentialSecret missing", func() {
```
shenoypritika, line 271: Is this for the DB Credentials? If yes, could we rename this for better clarity

shenoypritika (changes requested):

api/v1alpha1/webhook_suite_test.go
```diff
@@ -126,7 +129,338 @@ var _ = BeforeSuite(func() {
...
+	})
+})
+
+var _ = Describe("Webhooks: Time Machine missing (Open-source)", func() {
+	It("Testing webhooks", func() {
+
+		database := &Database{
+			ObjectMeta: metav1.ObjectMeta{
+				// need to provide a unique name to avoid getting " "db" already exists"
+				// this is because the previous test was a happy case and a db CR was successfully created
```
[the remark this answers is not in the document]
akshmish: Hey @shenoypritika Good suggestion. Just noticed that 8 of these 10 tests are NOT actually creating any resource as we are performing the negative testing i.e. assert the error returned by the Admission Webhooks. In the case of reject from the Webhooks, K8s will likely not create any resource so nothing to be cleaned up with 'AfterEach()'.

We do have 2 tests that successfully create a resource and those are being deleted by the AfterSuite(), so I am not sure running AfterEach() after ever step would help here, in fact it may make tests run slower. Generally we use Before/AfterEach() if we need isolation, but we do not have any such requirement here.

But I am ok to add AfterEach() if you can see a strong reason otherwise. Let me know.

api/v1alpha1/database_types.go
```diff
@@ -88,7 +88,7 @@ type Instance struct {
 	// Name(s) of the database(s) to be provisiond inside the database instance
 	// default [ "database_one", "database_two", "database_three" ]
 	// +optional
-	DatabaseNames []string `json:"databaseNames"`
+	DatabaseNames *[]string `json:"databaseNames"`
```
manavrajvanshi, line 91: We can remove the pointer from here, slices are inherently reference types.

api/v1alpha1/database_webhook.go
```diff
@@ -109,82 +115,93 @@ func (r *Database) Default() {
...
 var _ webhook.Validator = &Database{}
 
-func ndbServerSpecValidatorForCreate(r *Database, allErrs field.ErrorList, ndbPath *field.Path) field.ErrorList {
+func ndbServerSpecValidatorForCreate(ndb *NDB, allErrs field.ErrorList, ndbPath *field.Path) field.ErrorList {
 	databaselog.Info("Entering ndbServerSpecValidatorForCreate...")
 
-	if r.Spec.NDB == (NDB{}) {
-		allErrs = append(allErrs, field.Invalid(ndbPath, r.Spec.NDB, "NDB spec field must not be empty"))
+	if ndb == &(NDB{}) {
+		allErrs = append(allErrs, field.Invalid(ndbPath, ndb, "NDB spec field must not be empty"))
```
manavrajvanshi, line 122: I don't think we'll ever hit this case, because: the address of `(NDB{})` will not be known to the user. Would it be better to check `*ndb == (NDB{})` instead?

api/v1alpha1/database_webhook.go
```diff
@@ -109,82 +115,93 @@ func (r *Database) Default() {
...
 }
 
-func instanceSpecValidatorForCreate(r *Database, allErrs field.ErrorList, instancePath *field.Path) field.ErrorList {
+func instanceSpecValidatorForCreate(instance *Instance, allErrs field.ErrorList, instancePath *field.Path) field.ErrorList {
 	databaselog.Info("Entering instanceSpecValidatorForCreate...")
 
-	if r.Spec.Instance.DatabaseInstanceName == "" {
-		allErrs = append(allErrs, field.Invalid(instancePath.Child("databaseInstanceName"), r.Spec.Instance.DatabaseInstanceName, "A unique Database Instance Name must be specified"))
+	databaselog.Info("Logging the Instance details inside validator method", "databaseInstance", instance)
+
```
manavrajvanshi, line 145: We can combine these logs

api/v1alpha1/database_webhook.go
```diff
@@ -109,82 +115,93 @@ func (r *Database) Default() {
...
 
-	if r.Spec.Instance.Size < 10 {
-		allErrs = append(allErrs, field.Invalid(instancePath.Child("size"), r.Spec.Instance.Size, "Initial Database size must be 10 GBs or more"))
+	if instance.Size < 10 {
+		allErrs = append(allErrs, field.Invalid(instancePath.Child("size"), instance.Size, "Initial Database size must be specified with a value 10 GBs or more"))
 	}
 
-	if r.Spec.Instance.CredentialSecret == "" {
-		allErrs = append(allErrs, field.Invalid(instancePath.Child("credentialSecret"), r.Spec.Instance.CredentialSecret, "CredentialSecret must be provided"))
+	databaselog.Info("CredentialSecret", "CredentialSecret", instance.CredentialSecret)
```
manavrajvanshi, line 155: Redundant, will be printed in the previous log, should be removed.

manavrajvanshi (approved):
Looks good with follow ups on:
1. Pointer for slice type (redundant, can be removed)
2. Few logs
3. Reverting your private docker registry

api/v1alpha1/webhook_suite_test.go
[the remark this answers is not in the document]
shenoypritika: I think adding AfterEach will make the test suite more extensible. In the future, if we add more we need not worry about resource name conflicts.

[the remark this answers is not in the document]
akshmish: Sure, as discussed the cleanup after every test makes the test suite slower to run, but I am ok with it and have no specific preference on that. Updated the code. Resolving the comment.

shenoypritika (approved):

PR #135, opened 2023-07-31
NDBServer CRD + Controller

manavrajvanshi:
Adds NDBServer CRD and Controller with DB fetching

1. Tested existing DB Provisioning and app connectivity flows.
image

2. Tested NDB Server CRD creation and data fetching.
image

**Special notes for your reviewer**: None

**Release note**:

```release-note
None
```


api/v1alpha1/webhook_suite_test.go
```diff
@@ -141,335 +141,316 @@ var _ = AfterEach(func() {
 })
 
 var _ = Describe("Webhook Tests", func() {
-	Context("Functional Tests", func() {
```
yash-ntnx, line 144: most of these changes are coming because we have reverted the webhook txn?
and also because there is a new Database object in the ndb crd?

config/crd/bases/ndb.nutanix.com_ndbservers.yaml
```diff
@@ -35,13 +45,63 @@ spec:
           spec:
             description: NDBServerSpec defines the desired state of NDBServer
             properties:
-              foo:
-                description: Foo is an example field of NDBServer. Edit ndbserver_types.go
-                  to remove/update
+              credentialSecret:
                 type: string
+              server:
```
yash-ntnx, line 50: would this require changes in the helm repo too?

yash-ntnx (commented):
half way through the review. will finish today

api/v1alpha1/webhook_suite_test.go
manavrajvanshi: No, should've added a comment explaining these changes.
So basically the `Describe` and `It` blocks were redundant in these testcases so I removed them. Simply put, all the tests in this file were running in a separate context/suite. Now they run in a single suite.

config/crd/bases/ndb.nutanix.com_ndbservers.yaml
manavrajvanshi: Yes this will require us to add NDB CRD to the helm charts too. We should do that once the DB controller is refactored to make use of the NDB CR. WDYT?

controllers/common.go
```diff
@@ -0,0 +1,78 @@
...
+		} else {
+			errStatement = "NDB username or password cannot be empty"
+			err = fmt.Errorf("empty credentials")
+		}
+		log.Error(err, errStatement, "username empty", username == "", "password empty", password == "")
+		return
+	}
+	log.Info("Returning from common.getNDBCredentialsFromSecret")
+	return
+}
```
manavrajvanshi, line 78: Nothing new added here, just moved from Database controller helpers to a common file.

api/v1alpha1/ndbserver_types.go
```diff
@@ -20,26 +20,35 @@ import (
...
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	Status      string                           `json:"status"`
+	LastUpdated string                           `json:"lastUpdated"`
+	Databases   map[string]NDBServerDatabaseInfo `json:"databases"`
+	Counter     Counter                          `json:"counter"`
+}
+
+type Counter struct {
+	Database int `json:"database"`
```
akshmish, line 44: Easier to just keep "Counter" as an integer field inside NDBServerStatus?

manavrajvanshi: For now just putting the counter as an int field may be easier, yes!
But I purposely avoided it due to the following reason: Soon we'll need to add TM (and Snapshot) counters too, following the int field pattern we'll have 3 counters. The way reconcile works right now is to check if status and it's altered copy are deep equal (with the exception of the "Counter" key). We ignore the entire counter object from status deep equal check this way and would not have to do it for individual fields. 
Let me know how you feel about this.

controllers/ndbserver_controller.go
```diff
@@ -37,26 +45,89 @@ type NDBServerReconciler struct {
...
+		status.Databases = make(map[string]ndbv1alpha1.NDBServerDatabaseInfo)
+	}
+
+	// log.Info("NDBServer CR Status: " + util.ToString(status))
+
+	// 2. Verify credentials and connectivity
+	// Fetch credentials and check Authentication
+	username, password, caCert, err := getNDBCredentialsFromSecret(ctx, r.Client, ndbServer.Spec.CredentialSecret, req.Namespace)
+	ndbClient := ndb_client.NewNDBClient(username, password, ndbServer.Spec.Server, caCert, ndbServer.Spec.SkipCertificateVerification)
+	if err != nil {
```
yash-ntnx, line 87: we should check err before initializing `ndbClient`

controllers/ndbserver_controller_helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/nutanix-cloud-native/ndb-operator/ndb_api"
+	"github.com/nutanix-cloud-native/ndb-operator/ndb_client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// Fetches all databases from the NDB API and converts them to
+// NDBServerDatabaseInfo type object (to be consumed by the NDBServer CR)
+func getNDBServerDatabasesInfo(ctx context.Context, ndbClient *ndb_client.NDBClient) (databases []ndbv1alpha1.NDBServerDatabaseInfo, err error) {
+	log := log.FromContext(ctx)
+	log.Info("Entered ndbserver_controller_helpers.getNDBServerDatabasesInfo")
```
yash-ntnx, line 18: if we are retaining logs for such things then we should log something more meaningful (even if its an internal log)

controllers/ndbserver_controller_test.go
```diff
@@ -17,11 +17,86 @@ limitations under the License.
...
 
+		It("Should reconcile a NDBServer CR", func() {
+			By("Creating a NDBServer CR")
+			ndbServer := &v1alpha1.NDBServer{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      ndbSecretName,
+					Namespace: namespaceName,
+				},
+				Spec: v1alpha1.NDBServerSpec{
+					Server:           "https:111.222.333.444",
```
yash-ntnx, line 93: good thing we don't have ip address validation :)

ndb_api/auth.go
```diff
@@ -0,0 +1,65 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/nutanix-cloud-native/ndb-operator/ndb_client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func AuthValidate(ctx context.Context, ndbClient *ndb_client.NDBClient) (authValidateResponse AuthValidateResponse, err error) {
```
yash-ntnx, line 31: can you add some help text as to what this fn does

controllers/ndbserver_controller.go
```diff
@@ -37,26 +45,89 @@ type NDBServerReconciler struct {
...
+		return doNotRequeue()
+	}
 
-	// TODO(user): your logic here
+	// 4. Update the status if any changes are observed (excluding counter)
+	if !util.DeepEqualWithException(ndbServer.Status, *status, "Counter") {
+		log.Info("Status Changed, updating lastUpdated time")
+		status.LastUpdated = time.Now().Format(time.DateTime)
+	}
+	ndbServer.Status = *status
```
shenoypritika, line 117: Should this be inside the if block, since we want to update ndbServer.Status only if its changed?

controllers/ndbserver_controller.go
```diff
@@ -37,26 +45,89 @@ type NDBServerReconciler struct {
...
+	// 2. Verify credentials and connectivity
+	// Fetch credentials and check Authentication
+	username, password, caCert, err := getNDBCredentialsFromSecret(ctx, r.Client, ndbServer.Spec.CredentialSecret, req.Namespace)
+	ndbClient := ndb_client.NewNDBClient(username, password, ndbServer.Spec.Server, caCert, ndbServer.Spec.SkipCertificateVerification)
+	if err != nil {
+		log.Error(err, "Credential Error: error while fetching credentials from CredentialSecret", "secret name", ndbServer.Spec.CredentialSecret)
+		status.Status = common.NDB_CR_STATUS_CREDENTIAL_ERROR
+	} else {
+		authResponse, err := ndb_api.AuthValidate(ctx, ndbClient)
+		if err != nil || authResponse.Status != "success" {
```
shenoypritika, line 92: Could we make the "success" a constant

api/v1alpha1/ndbserver_types.go
shenoypritika: Could we rename "database_reconcile_counter" ? I got confused if this is the number of databases in NDB

shenoypritika (changes requested):

api/v1alpha1/ndbserver_types.go
manavrajvanshi: I don't want to repeat the word counter since it is already being used in the uber struct. I can however rename the counter to ReconcileCounter to address the confusion. WDYT?

controllers/common.go
```diff
@@ -0,0 +1,78 @@
...
+// requeue after a timeout. Put back on reconcile queue after a timeout and continue.
+func requeueWithTimeout(t int) (ctrl.Result, error) {
+	return ctrl.Result{RequeueAfter: time.Second * time.Duration(math.Abs(float64(t)))}, nil
+}
+
+// Returns the credentials(username, password and caCertificate) for NDB
+// Returns an error if reading the secret containing credentials fails
+func getNDBCredentialsFromSecret(ctx context.Context, k8sClient client.Client, name, namespace string) (username, password, caCert string, err error) {
+	log := ctrllog.FromContext(ctx)
+	log.Info("Entered common.getNDBCredentialsFromSecret")
```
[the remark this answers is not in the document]
manavrajvanshi: Remove the log?

controllers/common.go
```diff
@@ -0,0 +1,78 @@
...
+	secretDataMap, err := util.GetAllDataFromSecret(ctx, k8sClient, name, namespace)
+	if err != nil {
+		log.Error(err, "Error occured in util.GetAllDataFromSecret while fetching NDB secret", "Secret Name", name, "Namespace", namespace)
```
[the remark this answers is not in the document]
manavrajvanshi: `util.GetAllDataFromSecret` is a good catch, will remove. 
What specific error occurred during that will be logged by the first parameter to the log, more info on the secret are passed as key value pairs in the end of this log line. I can add an actionable statement to create the secret again and try again, is there anything else that you have in mind?

controllers/ndbserver_controller_test.go
manavrajvanshi: We should add that, thanks for pointing this out!

controllers/ndbserver_controller.go
manavrajvanshi: We still need to update the counter, hence this will be required on every reconcile.

controllers/ndbserver_controller_test.go
manavrajvanshi: We can't just limit this to IP address validation, there can be a fqdn / dns too right?

controllers/ndbserver_controller.go
```diff
@@ -37,26 +45,90 @@ type NDBServerReconciler struct {
...
+	// Fetch credentials and check Authentication
+	var ndbClient *ndb_client.NDBClient
+	username, password, caCert, err := getNDBCredentialsFromSecret(ctx, r.Client, ndbServer.Spec.CredentialSecret, req.Namespace)
+	if err != nil {
+		log.Error(err, "Credential Error: error while fetching credentials from CredentialSecret", "secret name", ndbServer.Spec.CredentialSecret)
+		status.Status = common.NDB_CR_STATUS_CREDENTIAL_ERROR
+	} else {
+		ndbClient = ndb_client.NewNDBClient(username, password, ndbServer.Spec.Server, caCert, ndbServer.Spec.SkipCertificateVerification)
+		authResponse, err := ndb_api.AuthValidate(ctx, ndbClient)
+		if err != nil || authResponse.Status != common.AUTH_RESPONSE_STATUS_SUCCESS {
```
akshmish, line 93: possible to check status code eg 200, instead of string compare?

controllers/ndbserver_controller.go
```diff
@@ -37,26 +45,90 @@ type NDBServerReconciler struct {
...
+			log.Error(err, "Authentication Error: Could not verify connectivity / auth credentials for NDB")
```
akshmish, line 94: "for NDB" -> "for NDB Server"?

controllers/ndbserver_controller_helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	// 1. Fetch dbs only if dbcounter is 0
+	if dbCounter == 0 {
+		log.Info("DbCounter 0, fetching databases (NDBServerDatabaseInfo)")
+		databases, err := getNDBServerDatabasesInfo(ctx, ndbClient)
+		if err != nil {
+			log.Error(err, "Error occurred while fetching databases (NDBServerDatabaseInfo)")
+			status.Status = common.NDB_CR_STATUS_ERROR
+		} else {
+			/* 2. TODO: Perform filtration on the databases associated with this NDB CR
+			databaseList := &ndbv1alpha1.DatabaseList{}
```
akshmish, line 62: Is this code working? If so, why are we not including it in the PR?

controllers/ndbserver_controller_helpers.go
```diff
@@ -0,0 +1,84 @@
...
+// 1. Checks and fetch data if dbcounter is zero (we fetch data only when counter hits 0).
+// 2. TODO: Filter and set the required list of databases (we only want to store the databases managed by the operator).
+// 3. Update the counter value.
+func getNDBServerStatus(ctx context.Context, status *ndbv1alpha1.NDBServerStatus, ndbClient *ndb_client.NDBClient) *ndbv1alpha1.NDBServerStatus {
+	log := log.FromContext(ctx)
+	log.Info("Entered ndbserver_controller_helpers.getNDBServerStatus")
+
+	dbCounter := status.ReconcileCounter.Database
+	// 1. Fetch dbs only if dbcounter is 0
+	if dbCounter == 0 {
```
akshmish, line 54: Trying to understand why we fetch data only when the counter reaches to 0?

ndb_api/auth.go
```diff
@@ -0,0 +1,67 @@
...
+// Returns the response along with an error (if any)
+func AuthValidate(ctx context.Context, ndbClient *ndb_client.NDBClient) (authValidateResponse AuthValidateResponse, err error) {
+	log := ctrllog.FromContext(ctx)
+	log.Info("Entered ndb_api.AuthValidate")
+	if ndbClient == nil {
+		err = errors.New("nil reference: received nil reference for ndbClient")
+		log.Error(err, "Received nil ndbClient reference")
+		return
+	}
+	res, err := ndbClient.Get("auth/validate")
```
akshmish, line 41: nit; we should consolidate all these hardcoded API paths into the constants file. We need to update other places as well when we fix that.

ndb_api/auth.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil || res == nil || res.StatusCode != http.StatusOK {
+		if err == nil {
```
akshmish, line 43: Do you need to conditionally log the res, err, status etc? We can just dump all three [res, err, status] if either of (err != nil || res == nil || res.StatusCode != http.StatusOK) is true. Would make the code simpler.

controllers/ndbserver_controller_helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	databases = make([]ndbv1alpha1.NDBServerDatabaseInfo, len(dbs))
+	for i, db := range dbs {
+		databaseInfo := ndbv1alpha1.NDBServerDatabaseInfo{
+			Name:   db.Name,
+			Id:     db.Id,
+			Status: db.Status,
+		}
+		if len(db.DatabaseNodes) > 0 {
+			databaseInfo.DBServerId = db.DatabaseNodes[0].DatabaseServerId
+			if len(db.DatabaseNodes[0].DbServer.IPAddresses) > 0 {
```
akshmish, line 34: Will you always pick 0th entry "db.DatabaseNodes[0]"? Just making sure.

controllers/ndbserver_controller.go
manavrajvanshi: I see what you're trying to say here, checking 200 would be a better approach too. But in the current implementation of ndb_api package methods the status code is not returned. We can make changes to that package to add a return value of the status too and make necessary re-factorings. I'm thinking of taking that up as a separate task/PR to not increase the scope of this one, WDYT?

controllers/ndbserver_controller.go
manavrajvanshi: Since these credentials are for the actual NDB (and not the NDB Server CR), I feel NDB fits more appropriately here. Let me know what you think..

controllers/ndbserver_controller_helpers.go
manavrajvanshi: No, this is not working currently.
The filtration logic will go here in the future when the Database CR is refactored to include NDB CR name, so we can filter the databases based on the id/ndb cr combination. Right now, we don't have a mechanism to filter the databases in the current mechanism that are being managed by the NDB CR. :D

controllers/ndbserver_controller_helpers.go
manavrajvanshi: The initial idea was to have the reconcile run every 'x' seconds (just to check if NDB connectivity is working fine etc) and update the databases, time machines and snapshots every 'dx', 'tx', 'sx' seconds. Since we're only dealing with databases initially, we simulate that behavior by maintaining this counter 'dbCounter'. Every time it reaches the `NDB_RECONCILE_DATABASE_COUNTER` (defined in constants, value: 4), it fetches the data for databases.

ndb_api/auth.go
manavrajvanshi: Cool suggestion! I'll include this in the ndb_api refactoring changes we discussed above!

ndb_api/auth.go
manavrajvanshi: These are specific errors that are being returned to the caller too, I'm not conditionally logging them. In order to provide the exact type of error to the caller, this if/else case might be necessary. WDYT?

controllers/ndbserver_controller_helpers.go
manavrajvanshi: Yes, for single instance databases we will always pick the 0th entry.

shenoypritika (approved):

controllers/common.go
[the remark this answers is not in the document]
yash-ntnx: are such logs generally needed? cause I see them in other functions too.

controllers/ndbserver_controller_test.go
yash-ntnx: yup, correct. so its not possible to validate the endpoint always.

yash-ntnx (approved):

controllers/ndbserver_controller_helpers.go
akshmish: Makes sense, please add a comment there for better context.

controllers/ndbserver_controller.go
akshmish: Sure, just leave a TODO comment on top of that method.

PR #140, opened 2023-08-16
Add Provisioning Operation Id to DB Status

shenoypritika:
- This PR adds provisioning operation ID field to DB status. This will help us track the operation status for provisioning operations in the upcoming PRs. 

image

### DIFF
--- a/api/v1alpha1/database_types.go
+++ b/api/v1alpha1/database_types.go
@@ -38,7 +38,7 @@ type DatabaseStatus struct {
 	Status                  string `json:"status"`
 	DatabaseServerId        string `json:"dbServerId"`
 	Type                    string `json:"type"`
-	ProvisioningOperationId string `json:"ProvisioningOperationId"`
+	ProvisioningOperationId string `json:"provisioningOperationId"`
 }
 
 // Database is the Schema for the databases API
--- a/api/v1alpha1/database_types.go
+++ b/api/v1alpha1/database_types.go
@@ -33,11 +33,12 @@ type DatabaseSpec struct {
 
 // DatabaseStatus defines the observed state of Database
 type DatabaseStatus struct {
-	IPAddress        string `json:"ipAddress"`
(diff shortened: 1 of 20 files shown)


mazin-s (approved):

api/v1alpha1/database_types.go
```diff
@@ -33,11 +33,12 @@ type DatabaseSpec struct {
...
-	Id               string `json:"id"`
-	Status           string `json:"status"`
-	DatabaseServerId string `json:"dbServerId"`
-	Type             string `json:"type"`
+	IPAddress               string `json:"ipAddress"`
+	Id                      string `json:"id"`
+	Status                  string `json:"status"`
+	DatabaseServerId        string `json:"dbServerId"`
+	Type                    string `json:"type"`
+	ProvisioningOperationId string `json:"ProvisioningOperationId"`
```
manavrajvanshi, line 41: Can you please change this to camelCase?

config/crd/bases/ndb.nutanix.com_databases.yaml
```diff
@@ -191,6 +193,7 @@ spec:
               type:
                 type: string
             required:
+            - ProvisioningOperationId
```
manavrajvanshi, line 196: Please change the json case to use camelCase and re-run `make generate//manifests` targets to update these values.

manavrajvanshi (changes requested):
Looks good, thanks! 
Please fix the case of the new status field to match json camelCase.

api/v1alpha1/database_types.go
shenoypritika: Done!

manavrajvanshi (approved):
LGTM, Thank you!

yash-ntnx (approved):

yash-ntnx:
changes look good. just want to check if there is something we should UT here. may not be any new test case but augment any of the existing cases. if not, thats ok.

PR #141, opened 2023-08-17
Automation mssql

mazin-s:
This PR refactors testing utilities to allow for modularity and ease of future testing. It includes postgres and mssql tests.

I need to fix an issue related to appConnectivity for MSSQL. This involves creating an appropriate docker image and pod to use (this will not effect my test functionality). I've included a log below of mssql-si_test.log. Note appConnectivity is failing due to the above issue. Also note, the log of the pg-si_test.log is the same.


automation/appConnectivity.go
```diff
@@ -0,0 +1,66 @@
...
+		return http.Response{}, fmt.Errorf("'kubectl port-forward %s %s:%d' failed! %v. ", podName, localPort, podTargetPort, err)
+	} else {
+		logger.Printf("kubectl port-forward %s %s:%d succesful.", podName, localPort, podTargetPort)
+	}
+
+	// Wait for a brief period to let port-forwarding start
+	time.Sleep(2 * time.Second)
+
+	// Verify the forwarded port by making an HTTP request
+	url := fmt.Sprintf("http://localhost:%s", localPort)
```
manavrajvanshi, line 46: Hey @mazin-s 
Why are we using a port-forward instead of services?

automation/provisioning.go
```diff
@@ -0,0 +1,295 @@
...
+
+// This function is called from the SetupSuite() function of all testsuites.
+// It loads environment variables, instantiate resources, waits for db to be ready, and pod to start.
+func ProvisioningTestSetup(ctx context.Context, st *SetupTypes, clientset *kubernetes.Clientset, v1alpha1ClientSet *clientsetv1alpha1.V1alpha1Client, t *testing.T) (err error) {
+	logger := GetLogger(ctx)
+	logger.Println("ProvisioningTestSetup() starting...")
+
+	// Nil check
+	if st == nil || clientset == nil || v1alpha1ClientSet == nil {
+		errMsg := "Error: ProvisioningTestSetup() starting ended! "
```
manavrajvanshi, line 31: Can we use a better statement here instead of `starting ended`?
Perhaps something like `nil parameters/arguments` or `initialization failed` etc?

manavrajvanshi (changes requested):
Minor comments and a question:
Why do we have 2 pg - si tests? Can we combine those into one? 

Also, what do you think about moving the files `appConnectivity.go`, `provisioning.go` and `setup.go` into a commons folder?

PR #155, opened 2023-09-26
NDB and DB Integration

manavrajvanshi:

- Integrates NDB and DB Controllers.
- Removes controller tests.
- Adds NDB CR to automation tests.

- App connectivity test did not pass because of rancher issue. 
- Provisioning test passed (Screenshot attached)
image

**Release note**:

```release-note
None
```


api/v1alpha1/database_webhook.go
```diff
@@ -153,6 +130,10 @@ func instanceSpecValidatorForCreate(instance *Instance, allErrs field.ErrorList,
 		allErrs = append(allErrs, field.Invalid(instancePath.Child("databaseInstanceName"), instance.DatabaseInstanceName, "A valid Database Instance Name must be specified"))
 	}
```
mazin-s, line 132: Does the clusterId have a specific syntax? Maybe we can add a webhook validation like we did with the NDB sever.

api/v1alpha1/database_types.go
```diff
@@ -27,7 +27,8 @@ import (
 
 // DatabaseSpec defines the desired state of Database
 type DatabaseSpec struct {
-	NDB      NDB      `json:"ndb"`
+	// +kubebuilder:validation:Required
```
akshmish, line 30: Is it possible to move this to webhooks as well?

api/v1alpha1/database_webhook.go
```diff
@@ -120,29 +120,6 @@ func (r *Database) Default() {
 
 var _ webhook.Validator = &Database{}
 
-func ndbServerSpecValidatorForCreate(ndb *NDB, allErrs field.ErrorList, ndbPath *field.Path) field.ErrorList {
```
akshmish, line 123: Has this been migrated to ndbserver_webhook.go, as a part of another PR before?

automation/setup.go
```diff
@@ -141,8 +168,7 @@ func ProvisioningTestTeardown(dbSecret, ndbSecret *corev1.Secret, database *ndbv
 
 	// Delete Database
 	if database != nil {
-		database.Spec.NDB.Server = os.Getenv("NDB-SERVER")
-		database.Spec.NDB.ClusterId = os.Getenv("NDB-CLUSTER-ID")
+		database.Spec.Instance.ClusterId = os.Getenv("NDB-CLUSTER-ID")
```
akshmish, line 171: Setting this env variable only specific for the automation tests?

api/v1alpha1/database_types.go
manavrajvanshi: No, since this is a required field irrespective of other values, annotation based validations are sufficient for this.

api/v1alpha1/database_webhook.go
manavrajvanshi: We are not introducing webhooks for NDBServer CRD as of right now. The only validation needed is `required` which is the default behavior right now.

automation/setup.go
manavrajvanshi: Yes.

akshmish:
Can you update the README file with the updated instructions for the usage?

manavrajvanshi:
Yes, I plan to do that once all the code changes are approved to avoid back and forth during reviews :)

common/constants.go
```diff
@@ -22,10 +22,10 @@ const (
 
 	DATABASE_CR_STATUS_DELETING     = "DELETING"
 	DATABASE_CR_STATUS_EMPTY        = ""
-	DATABASE_CR_STATUS_READY        = "READY"
+	DATABASE_CR_STATUS_NOT_FOUND    = "NOT FOUND"
 	DATABASE_CR_STATUS_PROVISIONING = "PROVISIONING"
-
-	DATABASE_DEFAULT_INSTANCE_NAME = "database_instance_name"
+	DATABASE_CR_STATUS_READY        = "READY"
+	DATABASE_CR_STATUS_WAITING      = "WAITING"
```
manavrajvanshi, line 28: Added for the transient state when a db provisioning has been requested on NDB and the NDBServer CR has not reconciled yet.

controllers/database_controller.go
```diff
@@ -76,7 +74,24 @@ func (r *DatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	log.Info("Database CR Status: " + util.ToString(database.Status))
 
-	NDBInfo := database.Spec.NDB
+	// Fetch the NDBServer resource from the namespace
+	ndbServer := &ndbv1alpha1.NDBServer{}
+	ndbNamespacedName := req.NamespacedName
```
shenoypritika, line 79: Can you check if this does a deep/shallow copy

controllers/database_reconciler_helpers.go
```diff
@@ -145,59 +146,18 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
-				r.recorder.Eventf(database, "Warning", EVENT_CR_STATUS_UPDATE_FAILED, "Error: %s. %s.", err.Error())
-				return err
-			}
-		}
-
-	}
-	return nil
-}
-
 // The handleSync function synchronizes the database CR's with the database instance in NDB
```
shenoypritika, line 149: Can you update this comment, as we are now syncing with the NDB CR rather than the NDB instance directly

controllers/database_reconciler_helpers.go
```diff
@@ -242,62 +201,76 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
 		// log.Info(fmt.Sprintf("Provisioning response from NDB: %+v", taskResponse))
 
 		log.Info("Setting database CR status to provisioning and id as " + taskResponse.EntityId)
```
shenoypritika, line 203: Please update this log

shenoypritika (changes requested):

controllers/database_reconciler_helpers.go
```diff
@@ -242,62 +201,76 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+	// any failure occurrs before reaching the ready state, the failure
+	// would not cause the deletion to block the terminal.
+	// Also, setup and create network services.
+	// [DELETING]
+	// Delete the database instance and the VM as per the finalizers
+	// [NOT FOUND]
+	// Record an event and then do not requeue since the resource has been deleted externally
+	// or was not found on NDB
+	switch databaseStatus.Status {
+	case common.DATABASE_CR_STATUS_READY:
```
shenoypritika, line 255: Thanks for tackling the finalizers

yash-ntnx: what are finalizers?

controllers/database_reconciler_helpers.go
```diff
@@ -242,62 +201,76 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+		return r.handleDelete(ctx, database, ndbClient)
+	case common.DATABASE_CR_STATUS_NOT_FOUND:
+		r.recorder.Eventf(database, "Warning", EVENT_EXTERNAL_DELETE, "Error: Resource not found on NDB")
 	default:
-		// Do Nothing
+		// No-Op
 	}
 
-	return doNotRequeue()
+	return requeueWithTimeout(15)
```
yash-ntnx, line 273: can we define this timeout (15) as a constant.

controllers/database_controller.go
manavrajvanshi: Good catch, thanks!
Just checked, Golang copies the value internally. If this was a reference type (pointer etc) then it would have copied the reference only (essentially a shallow copy) but since we're directly assigning a struct/field it copies it. 

Here's a little exercise I did to verify this:
image

api/v1alpha1/ndbserver_types.go
```diff
@@ -44,8 +44,8 @@ type ReconcileCounter struct {
 	Database int `json:"database"`
 }
 
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
+// +kubebuilder:object:root=true
```
yash-ntnx, line 47: just for my knowledge, what does this do?

controllers/database_reconciler_helpers.go
manavrajvanshi: @yash-ntnx 
Finalizers is a list of strings that is attached to the Custom Resource to handle the deletion of external resources owned by the CR. We leverage these finalizers to perform deletion of the Database Instance and the Database Server VM.

From the [Operator SDK docs](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#external-resources)- 
Sometimes external resources or resources that are not owned by a custom resource, those across namespaces for example, need to be cleaned up when the parent resource is deleted. In that case [Finalizers](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers) can be leveraged. A deletion request for an object with a finalizer becomes an update during which a deletion timestamp is set; the object is not deleted while the finalizer is present. The reconciliation loop of the custom resource’s controller will then need to check whether a the deletion timestamp is set, perform the external cleanup operation(s), then remove the finalizer to allow garbage collection of the object. Multiple finalizers may be present on an object, each with a key that should indicate what external resources require deletion by the controller.

api/v1alpha1/ndbserver_types.go
manavrajvanshi: This marker tells the Operator SDK's underlying Kubebuilder to treat this golang type struct as a Kubernetes Kind/resource. This marker generates the implementation of the Object interface that all Kubernetes `Kinds` are supposed to implement.

More here:
1. https://book.kubebuilder.io/cronjob-tutorial/new-api
2. https://book.kubebuilder.io/reference/markers/object

akshmish (approved):

shenoypritika (approved):
LGTM!

PR #160, opened 2023-10-12
Added operation tracking

manavrajvanshi:
It adds operation tracking for situations like instance creation and de-registration. The following have been done in this PR:
1. Changed the "PROVISIONING" state to "CREATING" (So it can be used for cloning too in the future).
2. Added support for operations API to query for the specific operations.
3. Updated DB CRD to store de-registration operation Id.
4. Added operation tracking during creation and de-registration.
5. Updated the add-api makefile target to copy the copyright text in new files.
6. Removed unused statuses.

Automation Test
image

**Release note**:

```release-note

```


controllers/database_reconciler_helpers.go
```diff
@@ -72,71 +72,71 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
 	if controllerutil.ContainsFinalizer(database, common.FINALIZER_DATABASE_INSTANCE) {
-		// Check if the database instance id (database.Status.Id) is present in the status
-		// If present, then make a deprovisionDatabase API call to NDB
-		// else proceed with removing finalizer as database instance provisioning wasn't successful earlier.
-		if database.Status.Id != "" {
-			infoStatement := "Deprovisioning database instance from NDB."
+		// Check if the deregistration operation id (database.Status.DeregistrationOperationId) is empty
+		// If so, then make a deprovisionDatabase API call to NDB
+		// else proceed check for the operation completion before removing finalizer.
+		deregistrationOperationId := database.Status.DeregistrationOperationId
```
mazin-s, line 78: Why does an empty DeregistrationOperationId denote deprovisioning the instance from NDB. I know we had this earlier as well, is there a particular reason for this. Is that how the NDB API handles it?

controllers/database_reconciler_helpers.go
```diff
@@ -157,7 +157,7 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
 	databaseStatus.Type = database.Spec.Instance.Type
 
 	// Provision the database if it has not been provisioned earlier
-	if databaseStatus.Status == "" || databaseStatus.Id == "" {
+	if databaseStatus.Status == "" && databaseStatus.Id == "" {
```
mazin-s, line 160: I find it confusing that when the database.Status.DeregistrationOperationId is empty we deregister the database from NDB. But when status is empty it means we provision. What do you think?

controllers/database_reconciler_helpers.go
```diff
@@ -260,11 +269,15 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
 		}
-		r.setupConnectivity(ctx, database, req)
+		if databaseStatus.IPAddress != "" {
+			r.setupConnectivity(ctx, database, req)
+		}
 	case common.DATABASE_CR_STATUS_DELETING:
 		return r.handleDelete(ctx, database, ndbClient)
 	case common.DATABASE_CR_STATUS_NOT_FOUND:
 		r.recorder.Eventf(database, "Warning", EVENT_EXTERNAL_DELETE, "Error: Resource not found on NDB")
+	case common.DATABASE_CR_STATUS_CREATION_ERROR:
```
mazin-s, line 279: Interesting, previously when we tried provisioning and there was something wrong with the yaml or provisioning request I believe it kept requeing? Now you'lll have it stop, nice, tracking the operation id is powerful.

ndb_api/operation.go
```diff
@@ -0,0 +1,76 @@
...
+		if err == nil {
+			if res != nil {
+				err = fmt.Errorf("GET %s responded with %d", getOperationPath, res.StatusCode)
+			} else {
+				err = fmt.Errorf("GET %s responded with a nil response", getOperationPath)
+			}
+		}
+		log.Error(err, "Error occurred fetching operation")
+		return
+	}
```
mazin-s, line 61: This if is a bit confusing, would it be better to seperate the different branches and saved the following string: "Error occurred fetching operation" in an accessible scope.

if err != nil
elif err == nil && res == nil
elif err == nil && res != nil && res.StatusCode != http.StatusOk

ndb_api/operation.go
```diff
@@ -0,0 +1,76 @@
...
+	log.Info(getOperationPath, "HTTP status code", res.StatusCode)
+	body, err := io.ReadAll(res.Body)
+	defer res.Body.Close()
+	if err != nil {
```
mazin-s, line 65: Could be add the id here?

ndb_api/operation_helpers.go
```diff
@@ -0,0 +1,28 @@
...
+limitations under the License.
+*/
+
+package ndb_api
+
+// Returns a boolean indicating if an operation has reached a terminal state
+func HasOperationFailed(o OperationResponse) bool {
+	status := o.Status
+	return status == "2" || status == "3" || status == "4"
+}
```
mazin-s, line 23: Could you add comments indicating what "2", "3", and "4" mean. I understand they are failures, but might be helpful to distinguish them for future developers.

ndb_api/operation_helpers.go
```diff
@@ -0,0 +1,28 @@
...
+
+// Returns a boolean indicating if an operation was successful
+func HasOperationPassed(o OperationResponse) bool {
+	return o.Status == "5"
+}
```
mazin-s, line 28: Same here.

mazin-s (commented):
Left some minor comments. Operation id tracking is great!

ndb_api/operation_helpers.go
```diff
@@ -0,0 +1,28 @@
...
+package ndb_api
+
+// Returns a boolean indicating if an operation has reached a terminal state
+func HasOperationFailed(o OperationResponse) bool {
+	status := o.Status
+	return status == "2" || status == "3" || status == "4"
+}
+
+// Returns a boolean indicating if an operation was successful
+func HasOperationPassed(o OperationResponse) bool {
```
shenoypritika, line 26: Can just have one function, say getOperationStatus that returns an enum PASS (0) for FAIL(1)?

controllers/database_reconciler_helpers.go
manavrajvanshi: We only reach the `handleDelete` function when the user has triggered a deletion of the Database Custom Resource. Now since we know that this resource has to be deleted, we make a deletion/de-registration API call to NDB, and store the deletion operation Id in the database status. 

Since K8s reconcile works in a loop (at least for our operator), this code path will be executed over and over again till the time the resource no longer exists (is deleted). To avoid multiple DELETE /databases/db-id API calls, we check if the deprovisioningOperationId was set in an earlier reconcile (started deletion on NDB already in one of the previous reconcile cycles).

Let me know if this helps.

controllers/database_reconciler_helpers.go
```diff
@@ -72,71 +72,71 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
-					errStament := fmt.Sprintf("Deprovisioning database server request failed for id: %s", databaseServerId)
-					log.Error(err, errStament)
-					r.recorder.Eventf(database, "Warning", EVENT_DEPROVISIONING_FAILED, "Error: %s. %s", errStament, err.Error())
+			database.Status.DeregistrationOperationId = deregistrationOp.OperationId
+			if err := r.Status().Update(ctx, database); err != nil {
+				return requeueOnErr(err)
+			}
+		} else {
+			deregistrationOp, err := ndb_api.GetOperationById(ctx, ndbClient, deregistrationOperationId)
+			if err != nil || ndb_api.HasOperationFailed(deregistrationOp) {
```
shenoypritika, line 96: - Can we separate out the errors here?
- .GetOperationById failing does not mean "Deregistration Failed". It is independent of the actual operation status. So might be better to separate out the error check and error logging

controllers/database_reconciler_helpers.go
manavrajvanshi: The idea is - We store the operationIds of the operations that we perform on NDB (Provisioning and De-Provisioning) here in our case. 
So if an id is set, it means that the operation was submitted to NDB previously and we do not need to submit the operation. On the other hand, if the id is not set then we're supposed to perform the operation submission based on the current status of the database custom resource.

controllers/database_reconciler_helpers.go
manavrajvanshi: Yes, we can still introduce re-queuing for certain cases. But some states can be classified as non-recoverable error states where re-queuing might not help. We can add this framework later on.

shenoypritika (changes requested):

ndb_api/operation_helpers.go
manavrajvanshi: Yes we can, but I think this way I wouldn't have to delegate the status checking in the client code of this package. 
Ultimately, we're still checking for a `PASS` or a `FAIL` state where we decide our next steps/actions which is accomplished with less verbosity if we have these methods. Otherwise, the client of this package might get way too verbose. 
Also, if the status values change in the NDB API (highly unlikely but let us assume), then I might have to make change in only one function and not in all the places this function is called.

Please let me know if you'd want to discuss more on this.

automation/tests/pg_si_test/config/db-secret.yaml
```diff
@@ -5,4 +5,4 @@ metadata:
 type: Opaque
 stringData:
   password: <password-for-database>
-  ssh_public_key: <enter-ssh-public-key>
+  ssh_public_key: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCwyAhpllp2WwrUB1aO/0/DN5nIWNXJWQ3ybhuEG4U+kHl8xFFKnPOTDQtTK8UwByoSf6wqIfTr10ESAoHySOpxHk2gyVHVmUmRZ1WFiNR5tW3Q4qbq1qKpIVy1jH9ZRoTJwzg0J33W9W8SZzhM8Nj0nwuDqp6FS8ui7q9H3tgM+9bYYxETTg52NEw7jTVQx6KaZgG+p/8armoYPKh9DGhBYGY3oCmGiOYlm/phSlj3R63qghZIsBXKxeJDEs4cLolQ+9QYoRqqusdEGVCp7Ba/GtUPdBPYdTy+xuXGiALEpsCrqyUstxypHZVJEQfmqS8uy9UB8KFg2YepwhPgX1oN noname
```
manavrajvanshi, line 8: Added a random ssh public key for tests so we don't have to use our own key every time and remove it before checking in on github.

shenoypritika (approved):

automation/tests/pg_si_test/config/db-secret.yaml
mazin-s: Nice this is convenient.

ndb_api/operation.go
mazin-s: Gotcha makes sense, we can revisit this later then when we refactor the common code paths.

controllers/database_reconciler_helpers.go
```diff
@@ -214,18 +215,27 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
 	isUnderDeletion := !database.ObjectMeta.DeletionTimestamp.IsZero()
 	if isUnderDeletion {
 		databaseStatus.Status = common.DATABASE_CR_STATUS_DELETING
+	} else if databaseStatus.Status == common.DATABASE_CR_STATUS_CREATING {
+		creationOp, err := ndb_api.GetOperationById(ctx, ndbClient, databaseStatus.CreationOperationId)
+		opStatus := ndb_api.GetOperationStatus(creationOp)
```
yash-ntnx, line 220: chk err before extracting opStatus

controllers/database_reconciler_helpers.go
```diff
@@ -214,18 +215,27 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+		if err != nil {
+			r.recorder.Event(database, "Warning", EVENT_NDB_REQUEST_FAILED, "NDB API to fetch operation by id failed: "+err.Error())
```
yash-ntnx, line 222: add op-id to msg

controllers/database_reconciler_helpers.go
```diff
@@ -214,18 +215,27 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+		} else if opStatus == ndb_api.OPERATION_STATUS_FAILED {
+			err = fmt.Errorf("creation operation terminated with status: %s, message: %s", creationOp.Status, creationOp.Message)
```
yash-ntnx, line 224: include op-id

controllers/database_reconciler_helpers.go
```diff
@@ -260,11 +270,15 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
 				return r.addFinalizer(ctx, req, common.FINALIZER_DATABASE_SERVER, database)
 			}
 		}
-		r.setupConnectivity(ctx, database, req)
+		if databaseStatus.IPAddress != "" {
```
yash-ntnx, line 273: if ip is empty, should we log an error/event?

ndb_api/operation.go
```diff
@@ -0,0 +1,76 @@
...
+	"net/http"
+
+	"github.com/nutanix-cloud-native/ndb-operator/ndb_client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// Fetches and returns a operation by an Id
+func GetOperationById(ctx context.Context, ndbClient *ndb_client.NDBClient, id string) (operation OperationResponse, err error) {
+	log := ctrllog.FromContext(ctx)
+	log.Info("Entered ndb_api.GetOperationById", "operationId", id)
```
yash-ntnx, line 34: if not needed, pls remove.

ndb_api/operation.go
```diff
@@ -0,0 +1,76 @@
...
+	if ndbClient == nil {
+		err = errors.New("nil reference")
```
yash-ntnx, line 36: can you enhance the error msg in this object to indicate more details like mentioned in the log stmt in the next line

ndb_api/operation.go
```diff
@@ -0,0 +1,76 @@
...
+	res, err := ndbClient.Get(getOperationPath)
+	if err != nil || res == nil || res.StatusCode != http.StatusOK {
+		if err == nil {
+			if res != nil {
+				err = fmt.Errorf("GET %s responded with %d", getOperationPath, res.StatusCode)
+			} else {
+				err = fmt.Errorf("GET %s responded with a nil response", getOperationPath)
+			}
+		}
+		log.Error(err, "Error occurred fetching operation")
```
yash-ntnx, line 59: add op-id to the msg

ndb_api/operation.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		log.Error(err, "Error occurred reading response.Body in Get Operation by ID", "operation id", id)
+		return
+	}
+	err = json.Unmarshal(body, &operation)
+	if err != nil {
+		log.Error(err, "Error occurred trying to unmarshal.")
+		return
+	}
+	log.Info("Returning from ndb_api.GetOperationById")
```
yash-ntnx, line 74: remove if not needed

ndb_api/operation_helpers.go
```diff
@@ -0,0 +1,39 @@
...
+	// 3: SUSPENDED
+	// 4: FAILED
+	// 5: PASSED
+	switch o.Status {
+	case "2", "3", "4":
+		status = OPERATION_STATUS_FAILED
+	case "5":
+		status = OPERATION_STATUS_PASSED
+	default:
+		status = ""
```
yash-ntnx, line 36: what does an empty status mean for the callers? are we sure the callers will be ok if we return empty here?

controllers/database_reconciler_helpers.go
```diff
@@ -72,71 +72,72 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
-					errStament := fmt.Sprintf("Deprovisioning database server request failed for id: %s", databaseServerId)
-					log.Error(err, errStament)
-					r.recorder.Eventf(database, "Warning", EVENT_DEPROVISIONING_FAILED, "Error: %s. %s", errStament, err.Error())
+			database.Status.DeregistrationOperationId = deregistrationOp.OperationId
+			if err := r.Status().Update(ctx, database); err != nil {
+				return requeueOnErr(err)
+			}
+		} else {
+			deregistrationOp, err := ndb_api.GetOperationById(ctx, ndbClient, deregistrationOperationId)
+			opStatus := ndb_api.GetOperationStatus(deregistrationOp)
```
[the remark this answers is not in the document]
manavrajvanshi: Sure, I did that because it was being used in both the `elif` cases. Moving them to an uber `else` and eliminating the temp variable by using a switch.

controllers/database_reconciler_helpers.go
```diff
@@ -72,71 +72,72 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
+			if err != nil {
+				r.recorder.Event(database, "Warning", EVENT_NDB_REQUEST_FAILED, "NDB API to fetch operation by id failed: "+err.Error())
+			} else if opStatus == ndb_api.OPERATION_STATUS_FAILED {
+				err := fmt.Errorf("deregistration operation terminated with status: %s, message: %s", deregistrationOp.Status, deregistrationOp.Message)
+				log.Error(err, "Deregistration Failed")
+				r.recorder.Event(database, "Warning", "OPERATION FAILED", "Database creation operation failed with error: "+err.Error())
+			} else if opStatus == ndb_api.OPERATION_STATUS_PASSED {
+				r.recorder.Eventf(database, "Normal", EVENT_DEPROVISIONING_COMPLETED, "Database deprovisioned from NDB.")
+				log.Info("Removing Finalizer " + common.FINALIZER_DATABASE_INSTANCE)
+				controllerutil.RemoveFinalizer(database, common.FINALIZER_DATABASE_INSTANCE)
```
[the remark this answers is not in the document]
manavrajvanshi: @yash-ntnx
Finalizers is a list of strings that is attached to the Custom Resource to handle the deletion of external resources owned by the CR. We leverage these finalizers to perform deletion of the Database/Clone Instance and the Server VM.

From the [Operator SDK docs](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#external-resources)-
Sometimes external resources or resources that are not owned by a custom resource, those across namespaces for example, need to be cleaned up when the parent resource is deleted. In that case [Finalizers](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers) can be leveraged. A deletion request for an object with a finalizer becomes an update during which a deletion timestamp is set; the object is not deleted while the finalizer is present. The reconciliation loop of the custom resource’s controller will then need to check whether a the deletion timestamp is set, perform the external cleanup operation(s), then remove the finalizer to allow garbage collection of the object. Multiple finalizers may be present on an object, each with a key that should indicate what external resources require deletion by the controller.

In context of this change:
Once we have deleted the Database from NDB, removing the finalizers would mean that our external resource has been cleaned up. Kubernetes on seeing this CR marked for deletion and containing no finalizers, will ultimately delete it from the cluster too.

ndb_api/operation_helpers.go
manavrajvanshi: As discussed, an empty status means all of the other non-used statuses atm. We can add more statuses as and when required in the `case`.

mazin-s (approved):

yash-ntnx (approved):

PR #183, opened 2023-11-16
Added unit tests and reduced redundant code

manavrajvanshi:

1. Increases unit test coverage. (78%)
2. Removes redundant code.
3. Fixes static lint errors.
4. Adds a new Makefile target to give coverage numbers.
5. Introduces an interface for the rest client - `NDBClientHTTPInterface` to mock the responses from NDB instead of spinning up a new http server for every test.
6. Bumped up go version and workflows to `1.21.7` (can't use 1.22 as it's not compatible with controller-gen). Note setup-envtest@lateste uses go 1.22.
7. Removed samples reference in `config/manifests/kustomization`
8. Added e2e additional logging for setting up log environment failure case.

**How Has This Been Tested?**: `make test-coverage`

**Release note**:

```release-note
None
```


Makefile
```diff
@@ -114,6 +114,18 @@ vet: ## Run go vet against code.
...
+ut: manifests generate fmt vet envtest ## Run tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -t skip `go list ./... | grep -v "/automation" | grep -v "/sandbox" | grep -v "/controllers"` -coverprofile cover.out
+
+.PHONY: test-coverage
+test-coverage: manifests generate fmt vet envtest ## Run unit tests creating the output to report coverage
+	- rm -rf *.out  # Remove all coverage files if exists
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test `go list ./... | grep -v "/automation" | grep -v "/sandbox" | grep -v "/controllers"` -coverprofile=cover.out.tmp
+	cat cover.out.tmp | grep -v "zz_generated.deepcopy.go" > cover.out
+	go tool cover -func cover.out
+	rm cover.out.tmp
```
manavrajvanshi, line 127: Gives coverage per function and aggregates them.

common/util/secret.go
```diff
@@ -2,33 +2,13 @@ package util
...
-		if val, ok := secretData[key]; ok {
-			data = string(val)
-			log.Info("Returning from util.GetDataFromSecret")
-		} else {
-			err = fmt.Errorf(fmt.Sprintf("Key '%s' not present in the secret '%s' in namespace '%s'", key, name, namespace))
-		}
-	}
-	return
-}
-
```
manavrajvanshi, line 31: Unused.

ndb_api/clone.go
```diff
@@ -15,136 +15,53 @@ package ndb_api
...
-		log.Error(err, "Error occurred reading response.Body in GetAllClones")
-		return
-	}
-	err = json.Unmarshal(body, &clones)
-	if err != nil {
-		log.Error(err, "Error occurred trying to unmarshal.")
-		return
-	}
-	log.Info("Returning from ndb_api.GetAllClones")
 	return
```
manavrajvanshi, line 33: All the functions in ndb_api package were performing same nil checks for ndb client, empty params, http calls, data deserialization and error handling. Now, all the common logic has moved to a `sendRequest` function.

automation/util/setup.go
```diff
@@ -66,7 +65,7 @@ func LoadEnv(ctx context.Context) (err error) {
 	// Loading env variables
 	err = godotenv.Load("../../.env")
 	if err != nil {
-		return fmt.Errorf("Error: loadEnv() ended! %s", err)
+		return fmt.Errorf("error: loadEnv() ended! %s", err)
```
mazin-s, line 68: I remember we talked about this before. Is making error messages lowercase a operator convention, ndb convention, go convention or coding common convention? Curious.

Makefile
mazin-s: cool

automation/util/setup.go
```diff
@@ -170,23 +169,23 @@ func SetupTypeTemplates(ctx context.Context) (setupTypes *SetupTypes, err error)
 	}
 
 	// Create ndbSecret template automation.NDB_SECRET_PATH
-	ndbSecret := &v1.Secret{}
+	ndbSecret := &corev1.Secret{}
```
mazin-s, line 172: we have both a corev1 and a v1 that use: "k8s.io/api/core/v1". We should eliminate one.

common/util/filter_slice_test.go
```diff
@@ -0,0 +1,113 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := Filter(tt.args.items, tt.args.fn); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Filter() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestFindFirstInt(t *testing.T) {
```
mazin-s, line 63: I think we should rename the core function to FindFirstIntFromFilter, what do you think?

common/util/print_test.go
```diff
@@ -0,0 +1,66 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import "testing"
+
+func TestToString(t *testing.T) {
```
mazin-s, line 21: Nit: Rename to intToString? It's possible to convert other objects to their string representation, up to you.

automation/util/setup.go
manavrajvanshi: I was getting go-static lint error/warnings, so it was suggested by my IDE.

automation/util/setup.go
manavrajvanshi: Yes, that's what I did. Again, go-static check :)

common/util/filter_slice_test.go
manavrajvanshi: The core function is a generic function which is called `Filter` and can work with any data type.  I just tested it with an integer type here because golang did not allow me to use mixed types in tests.

common/util/print_test.go
manavrajvanshi: This is a test file, the underlying function is called "ToString", by go test convention we just append the name of the function to be tested with a `Test` prefix.

automation/util/setup.go
mazin-s: gotcha.

common/util/filter_slice_test.go
mazin-s: ah makes sense.

common/util/print_test.go
mazin-s: I know I was just convering if it would be better we name the core function to "inToString" but now that I think about it "ToString" is descriptive enough and is the convention in other languages.

controllers/database_reconciler_helpers.go
```diff
@@ -70,7 +70,7 @@ func (r *DatabaseReconciler) addFinalizer(ctx context.Context, req ctrl.Request,
 func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1alpha1.Database, ndbClient *ndb_client.NDBClient) (ctrl.Result, error) {
 	log := ctrllog.FromContext(ctx)
 	log.Info("Database CR is being deleted")
-	log.Info(database.ResourceVersion)
+	// log.Info(database.ResourceVersion)
```
mazin-s, line 73: Should we delete this log?

ndb_client/ndb_client.go
```diff
@@ -44,39 +51,36 @@ func NewNDBClient(username, password, url, caCert string, skipVerify bool) *NDBC
...
+	url := ndbClient.url + "/" + endpoint
+
+	var body io.Reader
+
+	if requestBody != nil {
+		// Serialize the request body to JSON.
+		payload, err := json.Marshal(requestBody)
+		if err != nil {
+			return nil, err
+		}
```
akshmish, line 65: Make sure this is logged either here (preferably) or at the caller

ndb_client/ndb_client.go
```diff
@@ -44,39 +51,36 @@ func NewNDBClient(username, password, url, caCert string, skipVerify bool) *NDBC
...
-	payload, _ := json.Marshal(body)
-	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(payload))
+	// Create a new HTTP request with the specified method and URL.
+	req, err := http.NewRequest(method, url, body)
 	if err != nil {
-		// fmt.Println(err)
 		return nil, err
 	}
+
+	// Set headers
```
akshmish, line 76: optional; setting headers itself can be a separate method.

Makefile
```diff
@@ -114,6 +114,18 @@ vet: ## Run go vet against code.
...
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test `go list ./... | grep -v "/automation"` -coverprofile cover.out
 
+.PHONY: ut
+ut: manifests generate fmt vet envtest ## Run tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -t skip `go list ./... | grep -v "/automation" | grep -v "/sandbox" | grep -v "/controllers"` -coverprofile cover.out
+
+.PHONY: test-coverage
+test-coverage: manifests generate fmt vet envtest ## Run unit tests creating the output to report coverage
+	- rm -rf *.out  # Remove all coverage files if exists
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test `go list ./... | grep -v "/automation" | grep -v "/sandbox" | grep -v "/controllers"` -coverprofile=cover.out.tmp
```
akshmish, line 124: optional; you can combine "grep -v" on multiple keywords together.

akshmish:
For some filenames there is camelCase and for some filenames there is "snake_case". We should finalize a single casing.

akshmish (approved):
Approved with a few comments.

controllers/database_reconciler_helpers.go
manavrajvanshi: Good catch, thanks! :)

ndb_client/ndb_client.go
manavrajvanshi: Yes absolutely, it is logged by the common caller for all the API calls (i.e. sendRequest function). Did not log within the ndbClient package because we do not have a reference to the context object here.

ndb_client/ndb_client.go
manavrajvanshi: We can do that once we have more headers to set, I think it should be fine for just 2 headers for now? What do you think?

shenoypritika (approved):
LGTM! Thanks

yash-ntnx (approved):

PR #232, opened 2026-04-16
Feature ERA-63189: Support for Postgres HA instance

shivaprasadmb:
Adds support for provisioning, managing, and cloning Postgres High Availability (HA) database instances using NDB Operator. Previously the operator only supported single-instance (SI) database provisioning. This PR extends the operator to support Patroni-based HA clusters with HAProxy load balancing across one or more Nutanix Prism Element (PE) clusters.

- Provisioning of HA instance ( Virtual IP created)
Screenshot 2026-04-20 at 4 46 16 PM
Screenshot 2026-04-20 at 4 46 30 PM

- Provisioning of HA instance (without Virtual IP creation)
Screenshot 2026-04-20 at 6 29 53 PM
Screenshot 2026-04-20 at 7 47 32 PM
Screenshot 2026-04-20 at 7 47 43 PM

- Provisioning of DB clone instance (source as HA instance)
Screenshot 2026-04-20 at 7 53 23 PM
Screenshot 2026-04-20 at 7 10 44 PM

- Provisioning SI instance (for existing compatibility)
Screenshot 2026-04-20 at 1 16 49 PM
Screenshot 2026-04-20 at 1 18 01 PM

- NDB returns info fields in /dpcs/{id} responses as double-serialized JSON strings. GetDPCInfo() implements two-pass deserialization to handle this — first attempting direct unmarshal, then JSON-string unmarshal
- Operation status 17 (PASSED_WITH_WARNING) is now treated as success alongside status 5 (PASSED), as NDB commonly returns this for HA provisioning operations
- The haConfig field is intentionally not supported in defaultsConfigMapRef since topology (VM names, node-to-cluster assignments, Patroni cluster name) is unique per deployment and cannot be meaningfully defaulted.
- The Virtual IP would be only created when the HAProxy VMs reside in same subnet range. To make the provisioning of virtual IP, the HAProxy VMs were deployed in same PE cluster.
- For non-virtual IP based HA instance, the k8s service object keeps all HAProxy IPs as list inside it. Whereas in virtual IP scenario it points service object to single IP.
- Cloning of HA database instance can only be done into SI instance type and not cloned HA type.

**Release note**:

```release-note
- NDB server prerequisites: 
     Multi-cluster should be enabled and available PE clusters should be registered inside NDB.
     Stretched-VLAN should be created if NDB has to support Virtual IP creation for HA instance.
     Should have specific Software profile for Postgres HA which included Patroni softwre.
- Add support for provisioning, deleting, and cloning Postgres High Availability (HA) instances.
- Introduces `haConfig` in the Database CR spec with per-node topology configuration,
- Once the HA instance is provisioned, it creates two K8s svc objects (one for write and one for read)
```


.github/workflows/build-dev.yaml
manavrajvanshi, line 1: Please remove this change from this PR if unrelated.

.github/workflows/helm-lint.yaml
manavrajvanshi, line 1: Please remove this change from this PR if unrelated.

.github/workflows/helm-test.yaml
manavrajvanshi, line 1: Please remove this change from this PR if unrelated.

api/v1alpha1/database_types.go
```diff
@@ -84,6 +87,60 @@ func init() {
...
+	WritePort int32 `json:"writePort,omitempty"`
+	// HAProxy read (replica) port. Defaults to 5001.
+	// +optional
+	// +kubebuilder:default=5001
+	ReadPort int32 `json:"readPort,omitempty"`
+	// ProvisionVirtualIP controls whether NDB provisions a Virtual IP for the HAProxy layer.
+	// Set to true only when all HAProxy VMs are on the same Nutanix cluster (non-stretched VLAN).
+	// Defaults to false; use false when HAProxy nodes span multiple PE clusters.
+	// +optional
+	ProvisionVirtualIP bool `json:"provisionVirtualIP,omitempty"`
```
manavrajvanshi, line 138: Can you think of a more generic way instead?
As I understand, Patroni is only used for Postgres HA. We need to design this api in a way that we also think about what would happen when we onboard other HA enabled engines?

This approach will be too tightly coupled with Postgres in my opinion. 

I really appreciate the way you have named it generically - `HAConfig`, but it is too specific to a postgres HA use case currently. How about we add another level of indirection based on the engine - postgres, mysql, etc.

What do you think?

api/v1alpha1/webhook_helpers.go
```diff
@@ -243,6 +243,53 @@ func (v *ProvisioningWebhookHandler) validateCreate(spec *DatabaseSpec, errors *
 		*errors = append(*errors, field.Invalid(instancePath.Child("additionalArguments"), instance.AdditionalArguments, err.Error()))
 	}
 
+	// HA config validation
+	if instance.HAConfig != nil {
+		haPath := instancePath.Child("haConfig")
+
```
manavrajvanshi, line 249: Continuing on the comment above - this will also be based on the engine type being provisioned (or cloned). 

You could have a generic interface called HAParams validator and have multiple implementations of it - PostgresHAParamsValidator, MySQLHAParamsValidator etc and have the logic defined there. Much easier to maintain and extend rather than having long if-elses in the same block for multiple engines.

api/v1alpha1/webhook_helpers.go
```diff
@@ -243,6 +243,53 @@ func (v *ProvisioningWebhookHandler) validateCreate(spec *DatabaseSpec, errors *
...
+			}
+			if node.ClusterId == "" && node.ClusterName == "" {
+				*errors = append(*errors, field.Invalid(nodePath, node, "either clusterId or clusterName must be provided for each HA node"))
+			}
+			if node.ClusterId != "" {
+				if err := util.ValidateUUID(node.ClusterId); err != nil {
+					*errors = append(*errors, field.Invalid(nodePath.Child("clusterId"), node.ClusterId, "clusterId must be a valid UUID"))
+				}
+			}
+			if node.NodeType == "database" && node.Role == "Primary" {
```
manavrajvanshi, line 283: See if we can make use of constants instead of magic strings.

config/manager/kustomization.yaml
```diff
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: ghcr.io/nutanix-cloud-native/ndb-operator/controller
-  newTag: v0.5.4
+  newTag: v0.5.1
```
manavrajvanshi, line 16: We don't want to regress to an older version, please revert this. :)

controller_adapters/database.go
```diff
@@ -232,3 +232,54 @@ func (d *Database) GetCloneSourceDBId() string {
...
+	src := d.Spec.Instance.HAConfig
+	if src == nil {
+		return nil
+	}
+	nodes := make([]ndb_api.HANodeConfig, len(src.Nodes))
+	for i, n := range src.Nodes {
+		fm := n.FailoverMode
+		if fm == "" {
+			fm = "Automatic"
+		}
```
manavrajvanshi, line 259: Should this be a part of the defaulting webhook instead?

controller_adapters/database.go
```diff
@@ -232,3 +232,54 @@ func (d *Database) GetCloneSourceDBId() string {
...
+	}
+	writePort := src.WritePort
+	if writePort == 0 {
+		writePort = 5000
+	}
+	readPort := src.ReadPort
+	if readPort == 0 {
+		readPort = 5001
+	}
```
manavrajvanshi, line 275: These too

controllers/database_reconciler_helpers.go
```diff
@@ -113,8 +114,52 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
+			// then poll the returned operation ID until complete before removing the finalizer.
+			if len(database.Status.DBServerDeletionOperationIds) == 0 {
+				// Start phase: fire the DPC deletion and store the operation ID.
+				opIds, err := ndb_api.DeprovisionHADatabaseServers(ctx, ndbClient, database.Status.DatabaseServerId)
+				if err != nil {
+					log.Error(err, "Failed to deprovision one or more HA database servers")
+					r.recorder.Eventf(database, "Warning", EVENT_DEREGISTRATION_FAILED, "Error: %s", err.Error())
+				}
+				database.Status.DBServerDeletionOperationIds = opIds
+				if err := r.Status().Update(ctx, database); err != nil {
```
manavrajvanshi, line 130: Can you manually test this behavior when the update fails after making API calls to deprovision DBServers?

I suspect that we will lose track of the operation IDs in the subsequent reconciles an dtrigger new deprovisioning requests (which will be redundant operations and fail eventually). 

It might be better to retry once and the fail.

controllers/database_reconciler_helpers.go
```diff
@@ -258,8 +303,58 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+	if database.Spec.Instance != nil && database.Spec.Instance.HAConfig != nil {
+		haCfg := database.Spec.Instance.HAConfig
+		writePort := haCfg.WritePort
+		if writePort == 0 {
+			writePort = 5000
+		}
+		readPort := haCfg.ReadPort
+		if readPort == 0 {
+			readPort = 5001
+		}
```
manavrajvanshi, line 318: This feels redundant, should this be in the webhooks instead?

controllers/database_reconciler_helpers.go
```diff
@@ -258,8 +303,58 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
+			log.Error(err, errStatement)
+			r.recorder.Eventf(database, "Warning", EVENT_SERVICE_SETUP_FAILED, "Error: %s.", errStatement, err.Error())
+			return
+		}
+		if err = r.setupEndpoints(ctx, database, writeNN, writeMeta, writePort); err != nil {
+			errStatement := "Failed to setup write endpoints for HA database"
+			log.Error(err, errStatement)
+			r.recorder.Eventf(database, "Warning", EVENT_ENDPOINT_SETUP_FAILED, "Error: %s. %s", errStatement, err.Error())
+			return
+		}
```
manavrajvanshi, line 334: Lets avoid redundancy, we create the `-svc` anyway, and if the DB is HA enabled, we create the other services. WDYT?

controllers/database_reconciler_helpers.go
```diff
@@ -355,17 +490,12 @@ func (r *DatabaseReconciler) setupEndpoints(ctx context.Context, database *ndbv1
...
-			for _, address := range subset.Addresses {
-				if address.IP == database.Status.IPAddress {
-					// IP has not changed, no need to update endpoint
-					return
-				}
-			}
+		// If endpoint exists, check if the addresses have changed.
+		// If changed, sync with the latest IPs from the database CR status.
+		if endpointAddressesInSync(foundEndpoint.Subsets, desiredAddresses) {
+			return
```
manavrajvanshi, line 496: Q: Could these also be engine dependent in the structure? Meaning - can they change based on the engine type?

controllers/ndbserver_controller_helpers.go
```diff
@@ -39,8 +40,33 @@ func getNDBServerDatabasesInfo(ctx context.Context, ndbClient *ndb_client.NDBCli
 		}
 		if len(db.DatabaseNodes) > 0 {
 			databaseInfo.DBServerId = db.DatabaseNodes[0].DatabaseServerId
-			if len(db.DatabaseNodes[0].DbServer.IPAddresses) > 0 {
+
+			for _, n := range db.DatabaseNodes {
+				log.V(1).Info("DatabaseNode", "db", db.Name, "node", n.Name, "serverName", n.DbServer.Name, "ips", n.DbServer.IPAddresses, "properties", n.Properties)
```
manavrajvanshi, line 45: Please add a comment explaining why the explicit `log.V(1)` was needed?

controllers/ndbserver_controller_helpers.go
```diff
@@ -39,8 +40,33 @@ func getNDBServerDatabasesInfo(ctx context.Context, ndbClient *ndb_client.NDBCli
...
+				if err != nil {
+					log.Error(err, "Failed to look up HAProxy IPs via dbservers API", "db", db.Name)
+				} else {
+					haProxyIPs = clusterHAProxyIPs
+				}
+			}
+
+			if len(haProxyIPs) > 0 {
+				databaseInfo.IPAddress = strings.Join(haProxyIPs, ",")
+				log.Info("HA database: using HAProxy IPs", "db", db.Name, "ips", databaseInfo.IPAddress)
```
manavrajvanshi, line 66: Again, HA proxy is engine specific. Let's try to make this more modular.

deploy/helm/crds/customresourcedefinition-databases.ndb.nutanix.com.yaml
```diff
@@ -182,14 +171,88 @@ spec:
                     type: string
                   databaseNames:
                     description: |-
-                      Name(s) of the database(s) to be provisioned inside the database instance
+                      Name(s) of the database(s) to be provisiond inside the database instance
```
manavrajvanshi, line 174: Fix this please.

ndb_api/clone_helpers.go
```diff
@@ -276,6 +273,9 @@ func appendLCMConfigDetailsToRequest(req *DatabaseCloneRequest, additionalArgume
 		}
 	}
 	if refreshDetailsCount == 3 {
+		if req.LcmConfig == nil {
+			req.LcmConfig = &LcmConfig{}
+		}
```
manavrajvanshi, line 278: Q: Should this be a part of your PR?

manavrajvanshi (changes requested):
Reviewed once, massive PR!
Please feel free to discuss.

api/v1alpha1/database_types.go
shivaprasadmb: Understood, if MySQL were onboarded in the future then it would use completely different fields such as group replication / InnoDB cluster, MySQL Router etc which is different from Patroni Cluster, HAProxy fields in Postgres.
The best approach would be to keep the generic fields in InstanceHAConfig, and add an optional sub-struct per engine:
```
type InstanceHAConfig struct {
    ClusterName           string           // generic
    EnableSynchronousMode bool             // generic
    Nodes                 []InstanceHANode // generic
    // Exactly one of these should be set, matching instance.type
    Postgres *PostgresHAConfig `json:"postgres,omitempty"`
    // Future: MySQL *MySQLHAConfig `json:"mysql,omitempty"`
}
type PostgresHAConfig struct {
    PatroniClusterName string
    WritePort          int32
    ReadPort           int32
    ProvisionVirtualIP bool
}
```

Should we go with this approach?

.github/workflows/build-dev.yaml
shivaprasadmb: The changes is added to make the github workflow efficient, it used to run multiple redundant jobs in one commit. Now its reduced with this change. I believe we should keep this.

.github/workflows/helm-lint.yaml
shivaprasadmb: We are bypassing the Helm lint job for this change because:

- The lint job will fail if the chart version remains the same.
- While we did make CRD changes in deploy/helm, we aren't updating the chart version yet.
- We will run the lint job when we raise the actual deploy/helm changes later.

.github/workflows/helm-test.yaml
shivaprasadmb: Same scenario as above, I beleive we need this change.

controllers/ndbserver_controller_helpers.go
shivaprasadmb: `log.V(1)` keeps this out of production logs: this fires once per database node on every NDBServer sync cycle, so at INFO level it would flood logs in any deployment with more than a handful of databases.

controllers/database_reconciler_helpers.go
shivaprasadmb: Sure, i have made few fixes for this logic.

1. Generic updateStatusWithRetry helper

Instead of a bare Status().Update() after a destructive API call, all such persists now go through a helper that tries twice before giving up:

Attempt 1 — apply the status change, try to update
If it fails (usually a 409 Conflict from a stale resourceVersion) — re-fetch the latest object, re-apply, try once more
If both fail — log a CRITICAL message and return the error
2. persistHADeletionOpIds — wrapper for HA DPC deletion (FINALIZER_DATABASE_SERVER)

Also added an early return on API call failure — previously the code fell through with opIds = nil, persisted an empty list, and guaranteed the next reconcile would loop.

3. persistDeregistrationOpId — wrapper for database deregistration (FINALIZER_INSTANCE)

Applied the same protection to the older DeregistrationOperationId path, which had the identical vulnerability (affecting both HA and non-HA databases).

deploy/helm/crds/customresourcedefinition-databases.ndb.nutanix.com.yaml
shivaprasadmb: Done!

ndb_api/clone_helpers.go
shivaprasadmb: This was needed for my test scenarios to provision DBs.
Without the nil guard, clones that specify only refresh details (but not expiry details) would panic at runtime.

api/v1alpha1/database_types.go
shivaprasadmb: I have implemented the new approach, I have regenerated CRDs and tested the changes.

shivaprasadmb: InstanceHAConfig loses PatroniClusterName, WritePort, ReadPort, ProvisionVirtualIP; gains Postgres *PostgresHAConfig. New PostgresHAConfig struct holds those four fields.

api/v1alpha1/webhook_helpers.go
shivaprasadmb: HA validation block replaced with a registry lookup + generic pre-checks + validator.Validate() dispatch. No more engine-specific if-else here.

api/v1alpha1/webhook_helpers.go
shivaprasadmb: This is fixed, I have added below constants in common/constants.go and used in above logic:
HA_NODE_TYPE_DATABASE = "database"
HA_NODE_TYPE_HAPROXY  = "haproxy"
HA_NODE_ROLE_PRIMARY  = "Primary"
HA_NODE_ROLE_SECONDARY = "Secondary"

config/manager/kustomization.yaml
shivaprasadmb: This is fixed, i kept the version as 0.5.5

controller_adapters/database.go
shivaprasadmb: Yep, its fixed. Below is the new code in api/v1alpha1/webhook_helpers.go
```
// Apply HA defaults when haConfig is present
	if spec.Instance.HAConfig != nil {
		for i := range spec.Instance.HAConfig.Nodes {
			if spec.Instance.HAConfig.Nodes[i].FailoverMode == "" {
				spec.Instance.HAConfig.Nodes[i].FailoverMode = common.HA_NODE_FAILOVER_MODE_AUTOMATIC
			}
		}
		if pg := spec.Instance.HAConfig.Postgres; pg != nil {
			if pg.WritePort == 0 {
				pg.WritePort = common.HA_PROXY_DEFAULT_WRITE_PORT
			}
			if pg.ReadPort == 0 {
				pg.ReadPort = common.HA_PROXY_DEFAULT_READ_PORT
			}
		}
	}
```

controller_adapters/database.go
shivaprasadmb: Yes, updated with the above new code in api/v1alpha1/webhook_helpers.go

controllers/database_reconciler_helpers.go
shivaprasadmb: Fixed!

controllers/database_reconciler_helpers.go
shivaprasadmb: Sure, i have added a new file called controllers/ha_connectivity.go
- HAServiceSpec — a plain struct: {NameSuffix string, Port int32} describing an extra service to create
- HAConnectivityManager interface with two methods:
     - PrimaryPort(haConfig) — returns the port for the main -svc
     - AdditionalServices(haConfig) — returns specs for engine-specific extra services
- haConnectivityManagers registry map — add a new engine with one line
- PostgresHAConnectivityManager — returns HAProxy write port as primary, and [{"-ro-svc", readPort}] as the additional service

controllers/database_reconciler_helpers.go
shivaprasadmb: Its fixed along with above comment related code changes.

controllers/ndbserver_controller_helpers.go
shivaprasadmb: This is fixed, below is summary:
controllers/ha_connectivity.go — added two new types alongside the existing HAConnectivityManager:

- HAIPResolver interface — single method ResolveIPs(ctx, ndbClient, db) ([]string, error). Any future engine (MySQL HA, MongoDB HA) implements this and registers in the map.
- haIPResolvers registry — maps DATABASE_TYPE_* → implementation. Currently "postgres" → PostgresHAIPResolver.
- PostgresHAIPResolver — owns the entire HAProxy IP discovery logic that was previously inline:
     - ResolveIPs first calls collectHAProxyIPs (property/name match), then falls back to  GetHAProxyIPsForCluster (DPC endpoint).
     - collectHAProxyIPs is now a private method on the struct rather than a package-level function.

.github/workflows/helm-lint.yaml
manavrajvanshi: How do we plan on doing that? Will we need to raise another PR to revert these changes when were close to the release?

api/v1alpha1/database_webhook.go
```diff
@@ -39,18 +39,24 @@ var databaselog = logf.Log.WithName("database-resource")
...
 
 var _ admission.CustomDefaulter = &DatabaseCustomDefaulter{}
 
 func (r *Database) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	// In controller-runtime v0.21.0+, you must explicitly set the defaulter and validator
 	// The For() method alone does not automatically detect these interfaces
+	// Use GetAPIReader (direct, non-cached client) instead of GetClient (cache-backed).
+	// The webhook only reads a ConfigMap on each admission event, so caching adds no value.
+	// A cache-backed client would set up a cluster-wide ConfigMap watch, which requires the
+	// 'watch' verb; GetAPIReader only needs 'get', which we already have.
```
manavrajvanshi: Nit: Might be helpful to add why requiring the 'watch' verb is less preferred over 'get'.

api/v1alpha1/ha_validators.go
```diff
@@ -0,0 +1,83 @@
...
+	v, ok := haValidators[dbType]
+	return v, ok
+}
+
+// PostgresHAParamsValidator validates Patroni and HAProxy specific fields for Postgres HA.
+// +kubebuilder:object:generate=false
+type PostgresHAParamsValidator struct{}
+
+func (v *PostgresHAParamsValidator) Validate(haConfig *InstanceHAConfig, haPath *field.Path, errors *field.ErrorList) {
+	pgPath := haPath.Child("postgres")
```
manavrajvanshi, line 56: Please add UTs

api/v1alpha1/webhook_helpers.go
```diff
@@ -243,6 +243,49 @@ func (v *ProvisioningWebhookHandler) validateCreate(spec *DatabaseSpec, errors *
 		*errors = append(*errors, field.Invalid(instancePath.Child("additionalArguments"), instance.AdditionalArguments, err.Error()))
 	}
 
+	// HA config validation
+	if instance.HAConfig != nil {
+		haPath := instancePath.Child("haConfig")
+
+		validator, supported := getHAValidator(instance.Type)
```
manavrajvanshi: nit: Please rename this to indicate that this is validating HA inputs, perhaps something like 'haValidator' ?

api/v1alpha1/ha_validators.go
```diff
@@ -0,0 +1,83 @@
...
+func getHAValidator(dbType string) (HAParamsValidator, bool) {
+	v, ok := haValidators[dbType]
+	return v, ok
+}
+
+// PostgresHAParamsValidator validates Patroni and HAProxy specific fields for Postgres HA.
+// +kubebuilder:object:generate=false
+type PostgresHAParamsValidator struct{}
+
+func (v *PostgresHAParamsValidator) Validate(haConfig *InstanceHAConfig, haPath *field.Path, errors *field.ErrorList) {
```
manavrajvanshi, line 55: Nit: Please add a comment listing all fields that are being validated.

controllers/database_reconciler_helpers.go
```diff
@@ -36,6 +37,50 @@ import (
...
+	log := ctrllog.FromContext(ctx)
+	applyFn()
+	if err := r.Status().Update(ctx, database); err == nil {
+		return nil
+	}
+	log.Info("Status update conflict, re-fetching and retrying once")
+	if fetchErr := r.Get(ctx, req.NamespacedName, database); fetchErr != nil {
+		return fetchErr
+	}
+	applyFn()
```
manavrajvanshi, line 59: Why do we need to call this again when it has already been called once within the same function call?

controllers/database_reconciler_helpers.go
```diff
@@ -36,6 +37,50 @@ import (
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// updateStatusWithRetry applies applyFn to database and persists the status subresource,
+// retrying exactly once on failure. On the retry it re-fetches the latest object first
+// (resolving the 409 Conflict / stale resourceVersion that causes most update failures),
+// then calls applyFn again before the second attempt.
+// applyFn must be idempotent — it will be called before each attempt.
+// If both attempts fail, criticalMsg is logged as an error before returning.
+func (r *DatabaseReconciler) updateStatusWithRetry(ctx context.Context, req ctrl.Request, database *ndbv1alpha1.Database, applyFn func(), criticalMsg string) error {
```
manavrajvanshi: Why do we need the req here? Can we not get the namespace from the database object here?

controllers/database_reconciler_helpers.go
```diff
@@ -113,8 +156,56 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
+				if err != nil {
+					log.Error(err, "Failed to fetch HA server deletion operation", "operationId", opId)
+					allDone = false
+					continue
+				}
+				switch ndb_api.GetOperationStatus(op) {
+				case ndb_api.OPERATION_STATUS_PASSED:
+					log.Info("HA server deletion operation completed", "operationId", opId)
+				case ndb_api.OPERATION_STATUS_FAILED:
+					log.Info("HA server deletion operation failed", "operationId", opId, "message", op.Message)
```
manavrajvanshi: Should we also set allDone to false if operation fails?

controllers/database_reconciler_helpers.go
```diff
@@ -113,8 +156,56 @@ func (r *DatabaseReconciler) handleDelete(ctx context.Context, database *ndbv1al
...
+				return requeueWithTimeout(common.DATABASE_RECONCILE_INTERVAL_SECONDS)
+			}
+
+			// Poll phase: check every operation; requeue until all are done.
+			allDone := true
+			for _, opId := range database.Status.DBServerDeletionOperationIds {
+				op, err := ndb_api.GetOperationById(ctx, ndbClient, opId)
+				if err != nil {
+					log.Error(err, "Failed to fetch HA server deletion operation", "operationId", opId)
+					allDone = false
```
manavrajvanshi: I think this functionality is better suited for instance_manager.go

controllers/database_reconciler_helpers.go
```diff
@@ -255,35 +346,59 @@ func (r *DatabaseReconciler) handleSync(ctx context.Context, database *ndbv1alph
...
 		r.recorder.Eventf(database, "Warning", EVENT_SERVICE_SETUP_FAILED, "Error: %s.", errStatement, err.Error())
 		return
 	}
-	err = r.setupEndpoints(ctx, database, commonNamespacedName, commonMetadata, targetPort)
-	if err != nil {
-		errStatement := "Failed to setup kubernetes endpoints for database custom resource"
+	if err = r.setupEndpoints(ctx, database, svcNN, svcMeta, primaryPort); err != nil {
+		errStatement := "Failed to setup kubernetes endpoints for database"
 		log.Error(err, errStatement)
 		r.recorder.Eventf(database, "Warning", EVENT_ENDPOINT_SETUP_FAILED, "Error: %s. %s", errStatement, err.Error())
```
manavrajvanshi: How about moving this into setupServices and setupEndpoints instead? Wouldn't that be a cleaner approach?

controllers/ha_connectivity.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (m *PostgresHAConnectivityManager) AdditionalServices(haConfig *ndbv1alpha1.InstanceHAConfig) []HAServiceSpec {
+	readPort := common.HA_PROXY_DEFAULT_READ_PORT
+	if pg := haConfig.Postgres; pg != nil && pg.ReadPort != 0 {
+		readPort = pg.ReadPort
+	}
+	return []HAServiceSpec{
+		{NameSuffix: "-ro-svc", Port: readPort},
+	}
+}
```
manavrajvanshi, line 134: Please add UTs for this file.

controllers/ndbserver_controller_helpers.go
manavrajvanshi: Okay, then please add a comment on how to enable viewing these logs.

controllers/ndbserver_controller_helpers.go
```diff
@@ -39,8 +40,29 @@ func getNDBServerDatabasesInfo(ctx context.Context, ndbClient *ndb_client.NDBCli
...
+					log.Error(resolveErr, "Failed to resolve HA IPs", "db", db.Name)
+				} else if len(haIPs) > 0 {
+					databaseInfo.IPAddress = strings.Join(haIPs, ",")
+					log.Info("HA database: using engine-resolved IPs", "db", db.Name, "ips", databaseInfo.IPAddress)
+				}
+			}
+
+			if databaseInfo.IPAddress == "" && len(db.DatabaseNodes[0].DbServer.IPAddresses) > 0 {
 				databaseInfo.IPAddress = db.DatabaseNodes[0].DbServer.IPAddresses[0]
+				log.Info("Using first node IP", "db", db.Name, "ip", databaseInfo.IPAddress)
```
manavrajvanshi, line 68: We can do this regardless of HA type, right?
Just iterate on the list of DB nodes and add them to the IP String?
If there's 1 or 3 Nodes, the string will be created accordingly.


What do you think?

manavrajvanshi (commented):
Left a couple of comments, let's close on these soon. Please feel free to discuss anything that doesn't make sense.

controllers/ndbserver_controller_helpers.go
manavrajvanshi: Discussed offline - this is absolutely needed. @shivaprasadmb will add a comment explaining - Why?

api/v1alpha1/ha_validators.go
shivaprasadmb: Sure, added UTs in api/v1alpha1/ha_validators_test.go:
TestGetHAValidator
TestPostgresHAParamsValidator_Validate

api/v1alpha1/webhook_helpers.go
shivaprasadmb: fixed, renamed validator to haValidator.

api/v1alpha1/ha_validators.go
shivaprasadmb: Done!

controllers/database_reconciler_helpers.go
shivaprasadmb: Added a comment in code.
The second applyFn() call is necessary because r.Get() replaces the entire database struct in place with fresh data from the API server — including overwriting the status field that first applyFn() does. Without calling it again, we'd be sending the old, unchanged status to the API server on the retry.

controllers/database_reconciler_helpers.go
shivaprasadmb: I have updated the code now.
We will still keep allDone to true for failed operation because:
Terminal state — NDB will never retry it. Setting false would loop forever. Unblock the finalizer but emit log.Error + Warning event so the user knows to check for orphaned VMs manually

controllers/database_reconciler_helpers.go
shivaprasadmb: New changes are done.
- InstanceManager.deleteDatabaseServer return type: (task *ndb_api.TaskInfoSummaryResponse, err error) → (done bool, err error) — done=true means "proceed to remove finalizer", done=false means "requeue"
- non-HA return type: done=true doesn't mean "the NDB operation completed" — it means "this method has done its part, the caller can proceed". For non-HA that's always true on the first call since it's fire-and-forget. For HA it becomes true only once all operations are in a terminal state.

Non-HA flow
```
handleDelete
  └── instanceManager.deleteDatabaseServer()
        └── haConfig == nil → skip HA block
        └── deleteDatabaseServer()  ← fires DELETE /dbservers/{id}
                                       logs/records any error internally
        └── return true, nil
  └── done=true, err=nil
  └── removes FINALIZER_DATABASE_SERVER immediately
```

HA flow — first reconcile (start phase)
```
handleDelete
  └── instanceManager.deleteDatabaseServer()
        └── haConfig != nil → enter HA block
        └── DBServerDeletionOperationIds == [] → start phase
        └── DeprovisionHADatabaseServers() → opIds = ["op-abc"]
        └── persistHADeletionOpIds() → status.DBServerDeletionOperationIds = ["op-abc"]
        └── return false, nil
  └── done=false, err=nil
  └── requeueWithTimeout() ← does NOT remove finalizer
```

HA flow — subsequent reconciles (poll phase)
```
handleDelete
  └── instanceManager.deleteDatabaseServer()
        └── haConfig != nil → enter HA block
        └── DBServerDeletionOperationIds == ["op-abc"] → poll phase
        └── GetOperationById("op-abc")
              PASSED  → allDone stays true
              FAILED  → log.Error + Warning event, allDone stays true (terminal)
              default → allDone = false
        └── return allDone, nil
  └── done=false → requeueWithTimeout (still running)
      done=true  → removes FINALIZER_DATABASE_SERVER
```

controllers/database_reconciler_helpers.go
shivaprasadmb: I would say lets keep it as it is because:
setupService and setupEndpoints would need to internally do the port determination, HA manager lookup, and the additional-services loop. They'd each independently become responsible for creating ALL services (primary + additional) for a database.
Problems:
- the HAConnectivityManager lookup, service name construction, and the additional-services loop would have to exist in both setupService and setupEndpoints independently. This leads to duplicate orchestration logic.
- currently setupService and setupEndpoints are clean, single-responsibility. Keeping HA loop here breaks the primitive design.

Improvements done:
- each service+endpoint pair is now one setupServiceAndEndpoints call instead of two separate calls with duplicated error handling. The function now reads as a clear list of "what services does this database get"
- dropped req ctrl.Request — namespace is now read directly from database.Namespace.

controllers/ha_connectivity.go
shivaprasadmb: Done, here's test names:
PostgresHAConnectivityManager.PrimaryPort
PostgresHAConnectivityManager.AdditionalServices
PostgresHAIPResolver.collectHAProxyIPs
PostgresHAIPResolver.ResolveIPs

current mechanism:
haConnectivityManagers contains Postgres, rejects unknown engines
haIPResolvers contains Postgres, rejects unknown engines

will add MySQL support for test cases in next release.

controllers/ndbserver_controller_helpers.go
shivaprasadmb: - This is needed because when we consider Postgres HA, databaseNodes[] contains only the database VMs, not the HAProxy nodes. 
- Joining all node IPs would expose the Postgres primaries/secondaries directly as connection endpoints, bypassing HAProxy — that breaks read/write splitting and failover. 
- The HAIPResolver is specifically needed to resolve HAProxy IPs, which are only discoverable via the DPC endpoint. 
- The fallback to databaseNodes[0] is correct for single-node non-HA databases.

manavrajvanshi (approved):

controllers/ndbserver_controller_helpers.go
shivaprasadmb: Done!

sasikanthmasini (approved):
LGTM!

